Issue #43, opened 2018-09-13
Provide (V)DOM API for HTML


trieloff:
See adobe/project-helix#305

kptdobe:
I started to create the Bootstrap Grayscale example based on what we did during the hackathon. See WIP https://github.com/kptdobe/helix-bootstrap-grayscale.

Even though the API is not there, I mimic the API with some simple methods and I not really if this is where we want to go. See https://github.com/kptdobe/helix-bootstrap-grayscale/blob/master/src/html.pre.js#L40-L56

Whatever is the API called, this collection of DOM moves, wraps and decorations is what we'll end up doing to re-format the DOM to something which is design / css friendly. And this is really cumbersome to do and feels.. completely wrong!

cc @cwoeltge @simonwex @trieloff @markusheer @tripodsan

trieloff:
It looks a bit like a person with a hammer went to put some screws into a wall. I guess some of the transitions (especially `wrap`) might be done more elegantly in HTL. 

Are there good language that we can use for the problem domain of DOM2DOM transformations?

kptdobe:
That's where I wanted to go: I think now that DOM2DOM transformations via APIs is completely unnatural and wrong. 
As you mention, it would be more elegant in HTL which would mean that each section has an HTL representation that you might overwrite or create your own sections at project level. But then the expected input would be a JSON-like object (section.img.src or section.heading.text), not hast. mdast is close but might not be good neither because we'll keep having checks on the children objects (is child x an img or heading level 2).
I think we should still research on a good way to represent the "content" after the md transformation.

trieloff:
Here is an example what an array of sections with enhanced metadata could look like: https://github.com/adobe/hypermedia-pipeline/blob/29bda05d6ff5841de85469f54a458e7b32f73df2/test/fixtures/sectionsmetadata.json

We could also think about having a more compressed mdast-like structure as mentioned in #24 

Or: we could allow mdast2htast transformations with custom handlers, i.e. something like this:

```javascript

const converter = require('mdast-converter');

// matches using css-like expressions
// handles using an htl template
converter.match('list item:has(a)', require('list-item.htl'));

return { content: { htast: converter.convert(content.mdast) }};
```

kptdobe:
I am currently refactoring my example to go in the "define sections as htl" direction. Let see where it goes.

trieloff:
Fixed the example. For Noah, I've suggested #50

kptdobe:
[...]

PR #53, opened 2018-09-19
VDOM Util

trieloff:
For #43


trieloff:
@kptdobe please have a look. 

If we merge this, we can remove `content.htast` and `content.html` (and probably `content.children`) as they are redundant with `content.document`, `content.document.innerHTML` and `content.document.children`.

kptdobe:
ok, I'll review this afternoon.

kptdobe:
I tried the code with the initial project (`hlx demo test`) and the document object contains only an "blank" location object (no children, no innerHTML...)

kptdobe:
It is the "JSONification" of the document object which does not work. A breakpoint in the `pre.js` shows me the content.document object full of stuff!

trieloff:
The basic motivation for this is outlined in @simonwex's issue adobe/project-helix#305:

> #### Mark
> Mark is using Helix to power the adobe.io website by outputting HTML
> Mark needs to restructure the content to efficiently produce attractive layouts without the adobe.io content authors having to have special formatting knowledge
> - When styling the content, he uses CSS selectors to identify content
> - When manipulating the content structure, he works with the DOM API

So the idea is to give Mark an API for manipulating HTML that he is familiar with, instead of forcing him to learn HTAST (as great as it might be with all the unist-utils) or any new content structure (that we haven't defined in more than a couple of months, see #24)

> we now have a huge content.document object which is a representation of the browser DOM, including all events (`onmousemove`…) which is completely useless.

It might be useless, but I'd argue there is little harm done and it can increase compatibility when the developer is introducing their favorite library that happens to access `onmousemove` without any guards.

> I still do not understand who is going to use things like `content.document.getElementsByTagName('h1')` and to do what... My assumption is that a web developer will want to work with HTL and a developer preparing the data or convertir the sections into something project specific might want to go more for a simplified version of the mdast, basically a json representation.

Below are a couple of examples of how I could see this play out. They all have in common that the developer does not want to touch anything that looks like Markdown, but is very familiar with the DOM.

### Super Simple

Mark is fine with the HTML content and just wants it printed as is:

```html
<body>
${content.document.innerHTML}
</body>
```

Easy.

### Wrapping top-level elements

This replaces the `it.children` iterator.
```html
<div data-sly-list="${content.document.childNodes}">${item.innerHTML}</div>
```

### Adjusting Headlines

This takes a little work in `pre.js`:

```javascript
// remove the first heading
const firstheading = content.document.querySelector('h1');
firstheading.parentNode.removeChild(firstheading);

// downgrade all remaining h1 to h2
content.document.getElementsByTagName('h1').map(heading => {
  heading.outerHTML = '<h2>' + heading.innerHTML + '</h2>';
});
```

```html
<body>
<h1 class="myheading">${content.meta.title}</h1>
${content.document.innerHTML}
</body>
```

In all examples, you could even prototype the code on the client, in the web developer console before moving it server-side.

simonwex:
This is much tidier than I thought it would be. You nailed the use-case. Nice work!

I'll try to use it for real at some point this week, but from my static inspection, it looks like a great candidate to merge.

trieloff:
@kptdobe @simonwex I'm putting this back in WIP, as I noticed that the PR would break the image responsifyer, which is working on HTAST. I will need to re-implement that on top of the new DOM.

trieloff:
Some more documentation to describe how custom sizes for responsive images can be created.

Issue #50, opened 2018-09-18
Layout Machine


trieloff:
@simonwex built a state machine for inferring the content type of a section based on the sequence of children of the section. 

I think this could be a great complement to the simple frequency-based approach taken in #47 

I've been looking for a way to express the same pattern matching logic with less code and stumbled upon https://github.com/mrijk/speculaas (documentation here https://mrijk.github.io/speculaas/api-reference.html), which is a Javascript implementation of Clojure Spec (https://clojure.org/guides/spec#_sequences)

It would allow us to specify sequences in a syntax like this:

```javascript
// read as regex: heading? image image image+ i.e. one optional heading and at least three images
s.cat(s.question('heading'), 'image', 'image', s.plus('image'));

// one optional heading and then paragraphs or lists
s.cat(s.question('heading'), s.plus(s.alt('paragraph', 'list')));
```

@kptdobe do you think this would be useful?

kptdobe:
Yes, I think it is worth trying this (even though the syntax is not fully obvious on the first look)

trieloff:
I think we can turn this into a more obvious syntax, without having to rely on that library, simply by using regular expressions.

PR #47, opened 2018-09-17
Extract Sections and set section-specific metadata properties

trieloff:
This is a WIP, waiting for remarkjs/remark-frontmatter#6 to get merged.

This PR:
- adds a new property: `content.sections`, which is an Array of `Section` nodes, with each section
  - having an optional `meta` property with section-specific metadata, if the section has been created through a YAML block
  - having an `intro` and`title` property, using the first paragraph and heading of the section
  - having an `image` property, pointing to the `url` of the first image
  - having a `types` array that describes the content that can be found inside the section, using following patterns:
    - `has-<type>`: for each type of content that occurs at least once in the section, e.g. `has-heading`
    - `is-<type>-only`: for sections that only have content of a single type, e.g. `is-image-only`
    - `is-<type-1>-<type-2>-<type3>`, `is-<type-1>-<type-2>`, and `is-<type-1>` for the top 3 most frequent types of children in the section. For instance a gallery with a heading and description would be `is-image-paragraph-heading`.
 - takes `content.meta` as an aggregation of all `section` `meta` objects (same behavior as before)
 - copies `intro` and `title` from the first `section` (same as before)
 - copies `image` from the first section that has an image

Fixes #42

PR #54, opened 2018-09-21
Section Pattern Matching Language inspired by RegExp

trieloff:
This is a WIP implementation of #50 that does not require developers to build state machines from hand or to learn a new API to match sequences of elements, because it is simply using Regular expressions.

For instance to identify a "Gallery", i.e. a section that has at least three images, and maybe a heading, use `^heading? image image image+$`.

To identify a section that has only lists or paragraphs (and maybe a heading in the beginning), use `^heading? (paragraph|list)+$`.

It is intended to work with #47 and uses the new `utils` namespace from #53, so that it can be called from `pre.js`

Still outstanding:
- [x] merge #47 
- [x] provide `ContentMatcher` class with `match(expression, typename)` and `process` methods.
- [x] allow `match` to take in expressions or matching functions
- [x] let `process` set the matching `types` array
- [x] update README

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ The `types` property is an array of string values that describes the type of the
 
 - `has-<type>`: for each type of content that occurs at least once in the section, e.g. has-heading
 - `is-<type>-only`: for sections that only have content of a single type, e.g. is-image-only
-- `is-<type-1>-<type-2>-<type3>`, `is-<type-1>-<type-2>`, and `is-<type-1>` for the top 3 most frequent types of children in the section. For instance a gallery with a heading and description would be `is-image-paragraph-heading`.
+- `is-<type-1>-<type-2>-<type3>`, `is-<type-1>-<type-2>`, and `is-<type-1>` for the top 3 most frequent types of children in the section. For instance a gallery with a heading and description would be `is-image-paragraph-heading`. You can infer additional types using [`utils.types`](#infer-content-types-with-utilstypes).
 
 Each section has additional content-derived metadata properties, in particular:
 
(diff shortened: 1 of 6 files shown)


kptdobe:
The needs we have to implement the Bootstrap Grayscale template (https://blackrockdigital.github.io/startbootstrap-grayscale/) are:  

*  `h1 + h2 + paragraph` -> masthead section
* `(h2 + img + p)` -> about section
* `(h4 + img + p)*` -> rows of a project section

We might be able to tweak the last one for now and do `(h4 + img + p)` -> project section. They alternate left and right, we could use odd and even for now.

Covering this would be a good first step. The bottom of the page (subscribe form and contact details) could be a second step. Some for the "rows" of a section.

trieloff:
I can't match different levels of heading for now.

tripodsan:
I don't think that many developers are fluent in regexp :-) 
what about a sequence of CSS selectors ala jQuery ?

(on top of my head)
`.h1 .h2 .h3 .h4; img`

trieloff:
CSS selectors cannot express what is needed for this easily.

1. as you are selecting the `section`, you need the [`#has()` relational pseudo-class](https://drafts.csswg.org/selectors-4/#relational) (still in draft, so most people will be unfamiliar)
2. for expressing sequences of elements you only have the [next-sibling combinator `+`](https://drafts.csswg.org/selectors-4/#adjacent-sibling-combinators) and [subsequent-sibling combinator `~`](https://drafts.csswg.org/selectors-4/#general-sibling-combinators)
3. for expressing choices, you have to rely on the [`matches()` pseudo-class](https://drafts.csswg.org/selectors-4/#matches) (still in draft, too)
4. none of that is implemented in a query engine that is easily accessible

The pattern expressions are basically @simonwex's finite state machines, and quite easy as far as expression languages go.

PR #59, opened 2018-09-27
Surrogate keys

### LINKED ISSUE #23: Set better cache-control headers
*@trieloff commented on Feb 23, 2018, 11:10 AM UTC:*

*   etags (based on revision)
*   Vary (don't)
*   [Surrogate Keys](https://docs.fastly.com/guides/purging/getting-started-with-surrogate-keys) (based on template function, so that all stale content can be invalidated)

*This issue was moved by [trieloff](https://github.com/trieloff) from [adobe/project-helix#44](https://github.com/adobe/project-helix/issues/44).*

trieloff:
Fixes #23 

Each URL that is fetched individually to render the content should be listed in `content.sources`.

Each URL in `content.sources` will then be sha256 hashed and added to `Surrogate-Keys`.

If you want to write a selective cache invalidator (👋 @tripodsan), go over each file in the commit event, and go over each name of the commit (e.g. branch or tag names when pushed), calculate the `raw.githubusercontent.com` URL and get the hash.

### DIFF
--- a/src/html/set-surrogate-key.js
+++ b/src/html/set-surrogate-key.js
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2018 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+const crypto = require('crypto');
+
+function key({ content, response }, { logger }) {
+  // somebody already set a surrogate key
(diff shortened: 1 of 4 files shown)


src/html/set-surrogate-key.js
```diff
@@ -0,0 +1,33 @@
...
+  // somebody already set a surrogate key
+  if (!(response && response.headers && response.headers['Surrogate-Key']) && (content && content.sources && Array.isArray(content.sources))) {
+    logger.debug('Setting Surrogate-Key header');
+    return {
+      response: {
+        headers: {
+          'Surrogate-Key': content.sources.map((uri) => {
+            const hash = crypto.createHash('sha256');
+            hash.update(uri);
+            return hash.digest('hex');
```
tripodsan: shoulnd't we truncate this a bit or encode with base64? the hex encoded sha256 has 64 characters, which yields to very long surrogate-keys. using base64 would make it 4 times smaller.

trieloff: @tripodsan base64 sounds good, truncating not so much.

tripodsan: you didn't change it to base64.... ;-)

trieloff: Doesn't mean I'm not open to the change.

PR #60, opened 2018-09-27
Skip the fetch step if there is already a body

### LINKED ISSUE #13: Skip the fetch step in the default pipeline, if an input has been provided
See adobe/helix-cli#18:

if `resource.body` is a non-empty string in the `fetch` step, just skip the `fetch` step.

trieloff:
fixes #13

- this could be useful for you, @rofe


rofe:
> this could be useful for you, @rofe

trieloff:
We can change the CORS headers, or we can inject the stuff you need in the pipeline.

PR #61, opened 2018-10-02
Tap Pipeline & Dump Context for Debugging

### LINKED ISSUE #14: Provide tap function in the pipeline builder to enable in-progress inspection
Calling `pipe.pre(mystep).tap(myfn)`

should ensure that `myfn` will be called before `mystep` is executed by the pipeline, but that the return value of `myfn` will not enter the pipeline.

This allows setting `myfn` as an observer to `mystep`, which can ease debugging and inspection.

trieloff:
This fixes #14 and provides an alternative to #19:

when running in development-mode (i.e. no OpenWhisk in sight), every step of pipeline processing will be dumped in the `debug` directory. This allows for very fine-grained inspection of the pipeline processing.

The new `Pipeline.tap` helper function allows the registration of effect-less observers or validators.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,7 +395,7 @@ In the example above, all sections that have a `heading` as the first child will
 
 When run in non-production, i.e. outside an OpenWhisk action, for example in `hlx up`, Pipeline Dumping is enabled. Pipeline Dumping allows developers to easily inspect the `Context` object of each step of the pipeline and can be used to debug pipeline functions and to generate realistic test cases.
 
-Each stage of the pipeline processing will create a file like `$PWD/debug/context_dump_34161BE5KuR0nuFDp/context-2018-9-2-14-18-5.635-step-2.json` inside the `debug` directory. These dumps will be removed when the `node` process ends, so that after stopping `hlx up` the `debug` directory will be clean again. The `-step-n` in the filename indicates the step in the pipeline that has been logged.
+Each stage of the pipeline processing will create a file like `$PWD/debug/context_dump_34161BE5KuR0nuFDp/context-20180902-1418-05.0635-step-2.json` inside the `debug` directory. These dumps will be removed when the `node` process ends, so that after stopping `hlx up` the `debug` directory will be clean again. The `-step-n` in the filename indicates the step in the pipeline that has been logged.
 
 A simple example might look like this:
 
--- a/README.md
+++ b/README.md
@@ -390,3 +390,215 @@ In the example above, all sections that have a `heading` as the first child will
 * `heading paragraph* image` – a `heading` followed by any number of `paragraph`s (also no paragraphs at all), followed by an `image`
 * `(paragraph|list)` – a `paragraph` or a `list`
 * `^heading (image paragraph)+$` – one `heading`, followed by pairs of `image` and `paragraph`, but at least one
+
+### Inspecting the Pipeline Context
+
+When run in non-production, i.e. outside an OpenWhisk action, for example in `hlx up`, Pipeline Dumping is enabled. Pipeline Dumping allows developers to easily inspect the `Context` object of each step of the pipeline and can be used to debug pipeline functions and to generate realistic test cases.
+
+Each stage of the pipeline processing will create a file like `$PWD/debug/context_dump_34161BE5KuR0nuFDp/context-2018-9-2-14-18-5.635-step-2.json` inside the `debug` directory. These dumps will be removed when the `node` process ends, so that after stopping `hlx up` the `debug` directory will be clean again. The `-step-n` in the filename indicates the step in the pipeline that has been logged.
+
+A simple example might look like this:
+
+Step 0:
+```json
+{}
+```
+
+Step 1:
+```json
+{
+  "request": {}
+}
+```
+
+Step 2:
+```json
+{
+  "request": {},
+  "content": {
+    "body": "---\ntemplate: Medium\n---\n\n# Bill, Welcome to the future\n> Project Helix\n\n## Let's talk about Project Helix\n![](./moscow/assets/IMG_0167.jpg)\n",
+    "sources": [
+      "https://raw.githubusercontent.com/trieloff/soupdemo/master/hello.md"
+    ]
+  }
+}
+```
+
+Step 3 (diff only):
+```diff
+@@ -1,6 +1,58 @@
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -390,3 +390,215 @@ In the example above, all sections that have a `heading` as the first child will
 * `heading paragraph* image` – a `heading` followed by any number of `paragraph`s (also no paragraphs at all), followed by an `image`
 * `(paragraph|list)` – a `paragraph` or a `list`
 * `^heading (image paragraph)+$` – one `heading`, followed by pairs of `image` and `paragraph`, but at least one
+
+### Inspecting the Pipeline Context
+
+When run in non-production, i.e. outside an OpenWhisk action, for example in `hlx up`, Pipeline Dumping is enabled. Pipeline Dumping allows developers to easily inspect the `Context` object of each step of the pipeline and can be used to debug pipeline functions and to generate realistic test cases.
+
+Each stage of the pipeline processing will create a file like `$PWD/debug/context_dump_34161BE5KuR0nuFDp/context-2018-9-2-14-18-5.635-step-2.json` inside the `debug` directory. These dumps will be removed when the `node` process ends, so that after stopping `hlx up` the `debug` directory will be clean again. The `-step-n` in the filename indicates the step in the pipeline that has been logged.
```
tripodsan: pad days,months, and hours. hint: `String.padStart(num, char)`

src/defaults/html.pipe.js
```diff
@@ -31,8 +33,10 @@ const htmlpipe = (cont, payload, action) => {
   action.logger.log('debug', 'Constructing HTML Pipeline');
   const pipe = new Pipeline(action);
   pipe
+    .tap(dump).when(() => !production())
```
tripodsan: we have `pre`, `once`, `post`, why not call it `each` or `every` ?

src/utils/dump-context.js
```diff
@@ -0,0 +1,29 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+const tmp = require('tmp-promise');
+const path = require('path');
+const fs = require('fs-extra');
+
+fs.mkdirpSync(path.resolve(process.cwd(), 'debug'));
+const dumpdir = tmp.dir({ prefix: 'context_dump_', dir: path.resolve(process.cwd(), 'debug'), unsafeCleanup: true }).then(o => o.path);
```
tripodsan: why not put it in `logs/debug` ?

src/utils/dump-context.js
```diff
@@ -0,0 +1,29 @@
...
+
+function dump(context, _, index) {
+  return Promise.resolve(dumpdir).then((dir) => {
+    const now = new Date();
+    const dumppath = path.resolve(dir, `context-${now.getUTCFullYear()}-${now.getUTCMonth()}-${now.getUTCDay()}-${now.getUTCHours()}-${now.getUTCMinutes()}-${now.getUTCSeconds()}.${now.getUTCMilliseconds()}-step-${index}.json`);
```
tripodsan: - pad with 0s.
- i would make it more compact. remove `-` inside date/time, eg: `20180923-1434-123`

src/utils/dump-context.js
```diff
@@ -0,0 +1,29 @@
...
+    fs.writeJsonSync(dumppath, context, { spaces: 2 });
```
tripodsan: use async

src/utils/is-production.js
```diff
@@ -0,0 +1,20 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+/**
+ * Returns true when run inside an OpenWhisk action, false otherwise.
+ */
+function production() {
+  // eslint-disable-next-line no-underscore-dangle
+  return !!process.env.__OW_ACTIVATION_ID;
```
tripodsan: why not use `NODE_ENV` ?

tripodsan: also, what is the purpose of those dumps, if you cannot debug your deployed actions?

test/testHTML.js
```diff
@@ -170,4 +173,41 @@ describe('Testing HTML Pipeline', () => {
...
+      },
+    );
+
+    result.then((res) => {
+      assert.equal(201, res.statusCode);
+      assert.equal('text/plain', res.headers['Content-Type']);
+      assert.equal(res.headers['Surrogate-Key'], 'foobar');
+
+      dump({}, {}, -1).then((dir) => {
+        const outdir = path.resolve(dir, '..');
```
tripodsan: use `path.basename()`

package.json
```diff
@@ -52,6 +52,7 @@
     "retext": "^5.0.0",
     "retext-smartypants": "^3.0.1",
     "snyk": "^1.88.1",
+    "tmp-promise": "^1.0.5",
```
tripodsan: `tmp-promise` is not available on the docker image. suggest to use [tmp](https://www.npmjs.com/package/tmp)

src/defaults/html.pipe.js
trieloff: `tap` was the name suggested in https://www.amazon.com/dp/B0777JHD5M

src/utils/is-production.js
trieloff: Are we sure `NODE_ENV` is different in OpenWhisk and Petridish?

trieloff: > also, what is the purpose of those dumps, if you cannot debug your deployed actions?

See https://github.com/adobe/hypermedia-pipeline/pull/61#discussion-diff-222158166R36 the `tap` is only used `when` `!production`, i.e. in Petridish

package.json
trieloff: The docker image will pull in pipeline and its transitive dependencies, so that won't be an issue.

src/utils/is-production.js
trieloff: According to this post (http://heidloff.net/article/how-to-create-docker-actions-openwhisk-bluemix) `NODE_ENV` isn't forced to `production` in OpenWhisk, and not reliable.

PR #68, opened 2018-10-10
Parse the call stack for better log messages

### LINKED ISSUE #15: Improve "silly" logs in main pipeline function
Right now this is logging the entire `fn.toString()`, which leads to massively unreadable logs. Instead, it would be ideal to get the `CallSite` for the function that will be called/has been called.

trieloff:
fixes #15

### DIFF
--- a/src/pipeline.js
+++ b/src/pipeline.js
@@ -10,16 +10,20 @@
  * governing permissions and limitations under the License.
  */
 const _ = require('lodash/fp');
+const callsites = require('callsites');
 const Promise = require('bluebird');
 
+const noOp = () => {};
+
 const nopLogger = {
-  error: () => {},
-  warn: () => {},
-  info: () => {},
-  verbose: () => {},
-  debug: () => {},
-  silly: () => {},
-  log: () => {},
+  debug: noOp,
+  warn: noOp,
+  silly: noOp,
+  log: noOp,
+  info: noOp,
+  verbose: noOp,
+  error: noOp,
+  level: 'error',
 };
 
 /**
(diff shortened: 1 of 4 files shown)


src/pipeline.js
```diff
@@ -169,11 +171,42 @@ class Pipeline {
...
+    if (!f.alias && f.name) {
+      // eslint-disable-next-line no-param-reassign
+      f.alias = f.name;
+    } else if (!f.name && !f.alias) {
+      // eslint-disable-next-line no-param-reassign
+      f.alias = 'anonymous';
+    }
+
+    Error.captureStackTrace(f);
+    const [, current, injector, caller] = f.stack.split('\n').map(s => s.replace(/^.*at /, ''));
```
tripodsan: does this work in all locales? maybe better to use https://www.npmjs.com/package/traceback

src/pipeline.js
```diff
@@ -87,6 +87,7 @@ class Pipeline {
    * @returns {Pipeline} this
    */
   before(f) {
+    Pipeline.describe(f);
```
tripodsan: might not be so performant to capture the stack trace for every function. since it is only used for `loglevel === 'silly'`, you might check the loglevel here.

trieloff:
@tripodsan logs will looks like this:

```
silly: processing  {"function":"before:anonymous from Context.it (/code/hypermedia-pipeline/test/testPipeline.js:287:8)","index":0,"params":{}}
silly: received  {"function":"before:anonymous from Context.it (/code/hypermedia-pipeline/test/testPipeline.js:287:8)","result":{"foo":"bar"}}
silly: processing  {"function":"after:anonymous from Context.it (/code/hypermedia-pipeline/test/testPipeline.js:288:8)","index":1,"params":{"foo":"bar"}}
```

src/pipeline.js
trieloff: Good point.

src/pipeline.js
trieloff: Makes sense.

Issue #76, opened 2018-10-24
Examples from hlx demo (simple and full) fail to validate


trieloff:
As reported by @kptdobe.

PR #77, opened 2018-10-24
Relax Schemas to work with real-world Data

trieloff:
Ostensibly, this is a fix for #76, which is a bit too strict to handle the examples generated by `hlx demo`. As such, it can be merged and will fix a problem.

While going through the exercise of relaxing the schema, I had to document everything that happens in the pipeline right now, and there are a number of things that we should discuss. @tripodsan I'm going to leave review comments where appropriate, and the issues should then be filed against `petridish`


src/rawrequest.schema.json
```diff
@@ -20,12 +20,54 @@
...
+        },
         "params": {
           "type": "object",
           "additionalProperties": {
             "type": "string"
           },
           "properties": {
+            "req": {
+              "type": "string",
+              "description": "A JSON-encoded map of request parameters"
```
trieloff: @tripodsan do we still need this? In production, the value is never used and in development it just introduces an inconsistency.

src/request.schema.json
```diff
@@ -18,6 +18,59 @@
   "description": "The HTTP Request",
   "additionalProperties": false,
   "properties": {
+    "url": {
+      "type": "string",
+      "description": "The path and request parameters of the request URL",
```
trieloff: Right now, the `url` value is identical with `path` (because `path` does not chop off the query string). I think it's fine to leave it the way it is because it is consistent with node's handling: https://nodejs.org/api/http.html#http_message_url

src/request.schema.json
```diff
@@ -18,6 +18,59 @@
...
+    "resourcePath": {
+      "type": "string",
+      "description": "The resource path (without extension) that has been requested",
+      "examples": [
+        "/index"
+      ]
+    },
+    "path": {
+      "type": "string",
+      "description": "The path and request parameters of the request URL",
```
trieloff: This is inconsistent with Express: `path` should not include the query string: https://expressjs.com/en/4x/api.html#req.path

src/rawrequest.schema.json
tripodsan: we can drop this, since we now have `params`.

src/rawrequest.schema.json
```diff
@@ -20,12 +20,54 @@
...
+          "meta:enum": {
+            "get": "GET request",
+            "post": "POST request",
+            "put": "PUT request (not handled by Helix)",
+            "delete": "DELETE request (not handled by Helix)"
+          }
+        },
         "params": {
           "type": "object",
           "additionalProperties": {
```
tripodsan: would be nice to have a description here.

PR #82, opened 2018-10-31
processing html pipeline with non-existent content produces a 200 #81

tripodsan:
see #81

### DIFF
--- a/src/defaults/html.pipe.js
+++ b/src/defaults/html.pipe.js
@@ -19,6 +19,7 @@ const html = require('../html/make-html.js');
 const responsive = require('../html/responsify-images.js');
 const emit = require('../html/emit-html.js');
 const type = require('../html/set-content-type.js');
+const status = require('../html/set-status.js');
 const smartypants = require('../html/smartypants');
 const sections = require('../html/split-sections');
 const debug = require('../html/output-debug.js');
@@ -29,7 +30,7 @@ const validate = require('../utils/validate');
 
 /* eslint no-param-reassign: off */
 
-const htmlpipe = (cont, payload, action) => {
+const htmlpipe = async (cont, payload, action) => {
   action.logger = action.logger || log;
   action.logger.log('debug', 'Constructing HTML Pipeline');
   const pipe = new Pipeline(action);
@@ -47,6 +48,7 @@ const htmlpipe = (cont, payload, action) => {
     .before(emit)
     .once(cont)
     .after(type)
+    .after(status)
     .after(key)
     .after(debug);
 
--- a/src/html/set-status.js
+++ b/src/html/set-status.js
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2018 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+function setStatus({ content = {}, response = {}, error }, { logger }) {
+  // if a status is already default, keep it.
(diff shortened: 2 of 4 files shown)


src/defaults/html.pipe.js
```diff
@@ -29,7 +30,7 @@ const validate = require('../utils/validate');
 
 /* eslint no-param-reassign: off */
 
-const htmlpipe = (cont, payload, action) => {
+const htmlpipe = async (cont, payload, action) => {
```
kptdobe: Is this a required change ? side effects... ?

tripodsan: it was always returning a promise. but using `async` ensures that it really is one. but let me check again. I think I only did it for the tests.

kptdobe:
Did you tried these changes in production ? The VCL has some error handling, not sure if this plays together. See https://github.com/adobe/helix-cli/blob/master/layouts/fastly/helix.vcl#L551-L565

tripodsan:
> Did you tried these changes in production ? The VCL has some error handling, not sure if these play together. See https://github.com/adobe/helix-cli/blob/master/layouts/fastly/helix.vcl#L551-L565

the problem is, that even with no content, the HTL template produces some output (but not content). and the default status is a 200, set by `createActionResponse`.  so requesting anything will produce output.

src/html/set-status.js
```diff
@@ -0,0 +1,47 @@
...
+    };
+  }
+
+  // if there is an error, send a 500
+  if (error) {
+    logger.debug('payload.error -> 500');
+    return {
+      response: {
+        status: 500,
+        body: '',
```
trieloff: It would probably make sense to pass through the error message here, at least for Petridish

src/html/set-status.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+function setStatus({ content = {}, response = {}, error }, { logger }) {
+  // if a status is already default, keep it.
+  if (response.status) {
+    return {};
+  }
+
+  // if there is not content, send a 404
+  if (!content.html) {
```
trieloff: I don't think that's the right way to do it. There are many legitimate reasons why `content.html` might be absent (for instance because you are fetching JSON to from some other source).

Instead, I'd change the fetch action so that it guarantees `payload.error` to be set when fetching fails.

tripodsan: ok. I removed it.

src/pipeline.js
```diff
@@ -226,15 +253,25 @@ class Pipeline {
...
      * @param {Object} currContext Accumulated context
      * @param {pipelineFunction} currFunction Function that is currently "reduced"
+     * @param {number} index index of the function in the given array
      * @returns {Promise} Promise resolving to the new value of the accumulator
      */
-    const merge = (currContext, currFunction, index) => {
-      // copy the pipeline payload into a new object to avoid modifications
-      const mergedargs = _.merge({}, currContext);
+    const merge = async (currContext, currFunction, index) => {
+      const skip = (!currContext.error) === (!!currFunction.errorHandler);
```
trieloff: As I wrote in #85 I don't think this skipping magic is a good idea. Instead I would check if `context.errors!=null && this._errors.length> 1` (see above).

src/pipeline.js
```diff
@@ -175,7 +184,25 @@ class Pipeline {
...
+   */
+  error(f) {
+    this.describe(f);
+    if (!this._last) {
+      this._last = this._pres;
+    }
+    const wrapped = errorWrapper(f);
+    // ensure proper alias
+    wrapped.alias = f.alias;
+    this._last.push(wrapped);
```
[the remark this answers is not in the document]
tripodsan: no. otherwise you cannot interweave the errors into the list.

src/pipeline.js
tripodsan: the idea is, that the `error()` can be set like a `when()`:

```
before(fetch)
.error(recover_failed_fetch)
.once(...)
...
```

trieloff:
@tripodsan I can see your point, but it would be beneficial to describe how the `error` function works in `README.md`. The logic that all processing functions get skipped until the pipeline encounters an error handler can be beneficial, but only when people know about it.

tripodsan:
> The logic that all processing functions get skipped until the pipeline encounters an error handler can be beneficial, but only when people know about it

ok, I'll update the readme next week.. thanks.

Issue #87, opened 2018-11-02
Improve tests: remove position information


kptdobe:
Just discovered that for testing, we need to generate a full mdast that contains the position information (see example like https://github.com/adobe/hypermedia-pipeline/blob/master/test/fixtures/sections.json):

* 80% of the test validation json is useless
* impossible to create it manually which should be a pre-requisite for a good test scenario: you create a simple md file and the corresponding expected json and then you validate it against your code. This is impossible to do with the appropriate position info.
* outputting the JSON (https://github.com/adobe/hypermedia-pipeline/blob/master/test/markdown-utils.js#L25) on error to generate the test case is a really bad test practice. It is useful though when you want to test more complex md and generating the resulting mdast is really complex. But this should be an advanced final step in the test process.

Proposal here is to remove the position information in the json validation sample and thus on the fly in the pipeline output.

cc @trieloff

PR #88, opened 2018-11-02
Improve tests: remove position information #87

kptdobe:
PR for https://github.com/adobe/hypermedia-pipeline/issues/87

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,40 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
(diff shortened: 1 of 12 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,40 @@
+{
```
trieloff: Do we need this?

kptdobe: We have it in some other repo, like helix-cli. It is quite useful to not have to re-create it each time, especially the "Mocha Current File". I do not think we have a rule yet, it is IDE specific thus both having it or not could be argued. I am fine removing it.

PR #103, opened 2018-11-12
Fix for Regression in metadata extraction

trieloff:
See adobe/helix-cli#319

### DIFF
--- a/test/testHTML.js
+++ b/test/testHTML.js
@@ -152,8 +152,8 @@ describe('Testing HTML Pipeline', () => {
         // this is the main function (normally it would be the template function)
         // but we use it to assert that pre-processing has happened
         assert.equal(content.title, 'Helix - {{project.name}}');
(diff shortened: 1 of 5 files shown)


test/testHTML.js
```diff
@@ -146,6 +146,123 @@ describe('Testing HTML Pipeline', () => {
...
+        assert.equal(content.title, 'Helix - {{project.name}}');
+        assert.equal(content.image, './helix_logo.png');
+        assert.equal(content.intro, 'It works! {{project.name}} is up and running.');
+        // and return a different status code
+        return { response: { status: 201, body: content.html } };
+      },
+      {
+        content: {
+          body: `<!--
+~ Licensed to the Apache Software Foundation (ASF) under one or more
```
kptdobe: The markdown should be outside of the test file. It is unreadable as is...

Issue #106, opened 2018-11-15
Add XML pipeline


rofe:
Like the JSON pipeline, but assume that the output is a XML document.

Use case: implement a sitemap.xml

trieloff:
Just like the JSON pipeline, I don't think you would have an `xml.htl`, but instead write a straight `xml.js` with a main function.

rofe:
> Just like the JSON pipeline, I don't think you would have an `xml.htl`, but instead write a straight `xml.js` with a main function.
We should support both cases, I guess.

trieloff:
We do. The `xml.htl` would get compiled into `xml.js` anyway. `xml.htl` just sets the response body directly, so the additional validation step we discussed in #107 would be extra necessary then.

PR #107, opened 2018-11-15
Add XML pipeline #106

rofe:


trieloff:
Fixes #106 (you have to have the issue link in the description or a comment, GitHub does not pick it up from the title)

rofe:
Thanks for the feedback! I have a couple questions:
> * there is nothing that ensures the HTML will be valid

Would you expect the HTML to be generated, reusing the HTML pipe, and tested for XML validity in `testXML.js`? Or do you think XML checks should be applied as part of the processing?

> * the pipeline is operating on strings, not XML objects

What do you mean by that exactly? Should there be a pipeline step which creates a sort of DOM and store it e.g. in `content.xml`?

> I'd probably pick an XML library (ideally one that is already in our considerable dependency tree)...

I found [xml2js](https://www.npmjs.com/package/xml2js) in the package.json. Do you think that will do?

trieloff:
> I found xml2js in the package.json.
https://github.com/oozcitak/xmlbuilder-js/ (which is a dependency of xml2js) is probably better for what we need (JSON to XML).

trieloff:
>> there is nothing that ensures the HTML will be valid
> Would you expect the HTML to be generated, reusing the HTML pipe, and tested for XML validity in testXML.js? Or do you think XML checks should be applied as part of the processing?

Sorry for the confusion, I meant "ensures the XML is valid".

My expectation would be this (considering the `sitemap.xml` use case, but also things like RSS feed generation):

- the XML pipeline fetches and parses markdown
- the XML pipeline then calls your custom script (typically not an HTL template, but a plain JS function), which puts an object into `content.xml`
- the XML pipeline continues and
- turns the `content.xml` object into an valid XML `string` and puts it into the repose body
- sets the `Content-Type` header

Having an additional step that tries to parse the XML in response.body (if set directly in the user script) would make sense, too (in addition to the above)

rofe:
@trieloff: now the pipeline generates _some_ xml: `xmlroot` (Object) and `xml` (String) based on `mdast`
The question is, what would be a reasonable default xml output? Unlike HTML, I think for xml the output will almost always depend on the use case (atom feed, sitemap.xml etc), so I wonder if we should generate anything as part of the pipleine at all

(removed tests again for now)

rofe:
Or did I misunderstand your previous feedback, and it should be part of the test to generate _and_ validate xml?

rofe:
Incorporated changes as discussed and added test cases for individual steps as well as entire pipe.

Issue #108, opened 2018-11-15
Refactor: set defaults


trieloff:
The defaults for `secrets.*` are at the moment set per function, and sometimes documented in the schema. It would be better to have a new step that is injecting whatever `default` value has been defined in `secrets.schema.json` into the pipeline, so that documentation and implementation are always consistent and so that the same defaults will be used across all functions.

PR #111, opened 2018-11-20
Refactor defaults

trieloff:
This is an implementation of #108 

The `coerce` function uses the ability of the `ajv` JSON Schema validator to set defaults that are defined in the schema, but not in the passed object, and to coerce string values into more specific values. As a result, all configuration properties that are documented in the `secrets.schema.json` will be present in the `action` and have the proper type.

### DIFF
--- a/src/pipeline.js
+++ b/src/pipeline.js
@@ -84,7 +84,6 @@ class Pipeline {
 
     coerce(this._action);
 
-
     // function chain that was defined last. used for `when` and `unless`
     this._last = null;
     // functions that are executed first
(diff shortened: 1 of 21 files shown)


src/pipeline.js
```diff
@@ -81,6 +82,9 @@ class Pipeline {
 
     this._action.logger.debug('Creating pipeline');
 
+    coerce(this._action);
+
```
tripodsan: I think 1 empty line is enough :-)

src/utils/coerce-secrets.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+const ajv = require('./validator');
+
+async function coerce(action) {
+  const defaultsetter = await ajv(action.logger, { useDefaults: true, coerceTypes: true });
+  if (!action.secrets) {
+    /* eslint-disable no-param-reassign */
+    action.secrets = {};
+  }
+  if (action.secrets) {
```
tripodsan: this makes no sense.... is always `true`.

PR #120, opened 2018-11-26
Rewrite Links in HTML #25

rofe:
Implements #25

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,6 @@
     "unist-util-map": "^1.0.4",
     "unist-util-select": "^2.0.0",
     "uri-js": "^4.2.2",
-    "url": "^0.11.0",
     "winston": "^3.0.0",
     "xml2js": "^0.4.19",
     "xmlbuilder": "^10.1.1"
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "@adobe/openwhisk-loggly-wrapper": "^0.3.0",
     "ajv": "^6.5.4",
     "callsites": "^3.0.0",
+    "clone": "^2.1.2",
     "fs-extra": "^7.0.0",
     "hast-to-hyperscript": "^6.0.0",
     "hast-util-to-html": "^5.0.0",
@@ -65,6 +66,7 @@
     "unist-util-map": "^1.0.4",
     "unist-util-select": "^2.0.0",
     "uri-js": "^4.2.2",
+    "url": "^0.11.0",
     "winston": "^3.0.0",
     "xml2js": "^0.4.19",
     "xmlbuilder": "^10.1.1"
--- a/src/utils/link-handler.js
+++ b/src/utils/link-handler.js
@@ -18,8 +18,9 @@ const clone = require('clone');
 function link() {
   return function handler(h, node, children) {
     const n = clone(node);
-    if (!url.parse(n.url).protocol) {
-      n.url = n.url.replace(/\.md/, '.html');
+    const parts = url.parse(n.url);
+    if (!parts.protocol && parts.path) {
+      n.url = n.url.replace(parts.path, parts.path.replace(/\.md/, '.html'));
     }
     return fallback(h, n, children);
   };
--- a/src/utils/link-handler.js
+++ b/src/utils/link-handler.js
@@ -13,16 +13,15 @@
 const fallback = require('mdast-util-to-hast/lib/handlers/link');
 const uri = require('uri-js');
 
-
-function link() {
-  return function handler(h, node, children) {
+function link({ extension = 'html' } = {}) {
+  return function handler(h, node) {
     const n = Object.assign({}, node);
     const uriParts = uri.parse(n.url);
     if (!uriParts.scheme && uriParts.path) {
-      uriParts.path = uriParts.path.replace(/\.md/, '.html');
+      uriParts.path = uriParts.path.replace(/\.md$/, `.${extension}`);
       n.url = uri.serialize(uriParts);
     }
-    return fallback(h, n, children);
+    return fallback(h, n);
   };
 }
 
--- a/src/utils/link-handler.js
+++ b/src/utils/link-handler.js
@@ -10,16 +10,18 @@
  * governing permissions and limitations under the License.
  */
 
-function link({ LINK_TITLES }) {
-  return function handler(h, node) {
-    const props = {};
-    if (node.url) {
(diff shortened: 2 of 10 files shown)


src/utils/mdast-to-vdom.js
```diff
@@ -61,6 +62,7 @@ class VDOMTransformer {
         return true;
       });
     this._handlers.image = image(options);
+    this._handlers.embed = link(options);
```
trieloff: Typo: should be `this._handlers.link =…`

src/utils/link-handler.js
```diff
@@ -0,0 +1,26 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+function link({ LINK_TITLES }) {
+  return function handler(h, node) {
+    const props = {};
+    if (node.url) {
+      props.href = node.url.replace(/\.md/, '.html');
```
trieloff: It would be ideal to make the extension configurable and dependent on the extension in the request. This might require a bit of mangling as `VDOMTransformer` does not get the current request object, but I think it would increase reusability.

src/utils/link-handler.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+    if (LINK_TITLES) {
```
trieloff: I'd simply check if 'node.title' is set. Making the href the title does not seem like a good idea.

src/utils/link-handler.js
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+function link({ LINK_TITLES }) {
+  return function handler(h, node) {
+    const props = {};
+    if (node.url) {
```
trieloff: You are also rewriting external links, which does not seem like a good idea.

src/utils/mdast-to-vdom.js
trieloff: Btw. I'm unclear why your tests are passing, given that you register the wrong handler.

rofe:
Addressed several issues from initial commit, such as regressions (empty `<a>` tags), rewritten external links and several other problems. I also removed the title config (needs more thinking).

I tested this version with [Adobe-Developer-Site](https://github.com/adobe/Adobe-Developer-Site) and found no more regressions there. All tests are passing (except one which I commented out, see below):

TODOs:
- uncomment test "Custom Text Matcher with VDOM Nodes" in `testMdastToVDOM.js` - currently fails because the customized link nodes get overwritten by the new link handler

package.json
```diff
@@ -38,6 +38,7 @@
     "@adobe/openwhisk-loggly-wrapper": "^0.3.0",
     "ajv": "^6.5.4",
     "callsites": "^3.0.0",
+    "clone": "^2.1.2",
```
trieloff: Not needed: you can use `Object.assign({}, node)` instead.

src/utils/link-handler.js
```diff
@@ -10,16 +10,18 @@
...
+const fallback = require('mdast-util-to-hast/lib/handlers/link');
+const url = require('url');
+const clone = require('clone');
+
+
+function link() {
+  return function handler(h, node, children) {
+    const n = clone(node);
+    if (!url.parse(n.url).protocol) {
+      n.url = n.url.replace(/\.md/, '.html');
```
trieloff: Maybe just replace it in the path.

trieloff: You never know when someone registers the `.md` top level domain.

package.json
```diff
@@ -65,6 +66,7 @@
     "unist-util-map": "^1.0.4",
     "unist-util-select": "^2.0.0",
     "uri-js": "^4.2.2",
+    "url": "^0.11.0",
```
trieloff: Try to use `uri-js` instead of `url`.

package.json
rofe: I chose `clone` over `JSON.parse(JSON.stringify(obj))` or `Object.assign({}, obj)` because of potential problems with complex objects (such as `Date`)... but if in a node all we expect to be cloned are strings, numbers and simple objects, we can drop it again.

package.json
rofe: I chose `url` because `image-handler.js` was already using it.

src/utils/link-handler.js
rofe: good point

rofe: Fixed in https://github.com/adobe/hypermedia-pipeline/pull/120/commits/5f705848fe86fd2f1726d7b98ed1ce07734a5b0d

package.json
rofe: Fixed in https://github.com/adobe/hypermedia-pipeline/pull/120/commits/db14b2c6e01f917f1d116c4e419980003569abd2

package.json
rofe: Fixed in https://github.com/adobe/hypermedia-pipeline/pull/120/commits/f74f8695efaaaa0ca1ae80cdbbae6cccf81ab4bb

src/utils/link-handler.js
rofe: Fixed in https://github.com/adobe/hypermedia-pipeline/pull/120/commits/6386202b2e004517b064f2ed2991b8dded8739e9

src/utils/link-handler.js
rofe: Fixed in https://github.com/adobe/hypermedia-pipeline/pull/120/commits/6386202b2e004517b064f2ed2991b8dded8739e9

src/utils/mdast-to-vdom.js
rofe: Fixed in https://github.com/adobe/hypermedia-pipeline/pull/120/commits/6386202b2e004517b064f2ed2991b8dded8739e9

rofe:
last TODOs addressed with https://github.com/adobe/hypermedia-pipeline/pull/120/commits/d968da4d8dc99196ea87e7727ad7ba0a66959104

Issue #136, opened 2018-12-13
[sections] missing types of nested elements


kptdobe:
The `mdast` generated for a section with only one image looks like this:

```
{
  type: 'paragraph,
  children: [{
    type: 'image'
  }]
}
```

The resulting `types` contain only `has-paragraph` because the logic does not iterate through the children. 
More generally, all types of elements are encapsulated inside a parent paragraph, the `types` property does not compute meaningful info for most of the use cases.

Issue #112, opened 2018-11-21
List Comprehension


trieloff:
As an alternative to #71, list comprehension introduces lists as a critical top-level concept in the pipeline, with similar importance to sections.

Lists can be used in scenarios like
- navigation
- feature lists
- listicles
- product messaging

As a first step, list comprehension will simply look at nested content of list items and make the information available in HTML `class` attributes.

kptdobe:
At the moment, "information available" is NOT available as HTML `class` attributes but as section "types". And I think it should stay like this for now, up to the developer to dump the types as css classes or not. We start having a lot of types computed, we might not want to have all of them.
This may be part of the research here to determine a good format to expose this information for lists.

trieloff:
I think the likelihood of a developer iterating over sections is quite high, but for lists, I'd assume that most are simply passed through.

kptdobe:
I started to look into this.

kptdobe:
The various commits add the "types" info on the section's children (recursive for lists). It does not add the types as css class names but as data.types property on the section mdast nodes. This could be easily be transformed into css class names with the following code:

```js
visit(section, (child) => {
      if (child && child.data && child.data.types) {
        // assign the child types to the child className so that they get rendered
        child.data = Object.assign({
          hProperties: {
            className: child.data.types
          }
        }, child.data || {});
      }
    });
```

kptdobe:
I created https://github.com/adobe/helix-pipeline/issues/198 to follow up on this: document / investigate on the real usage of what is being done here.

trieloff:
I think it wouldn't hurt to directly inject it into the `data.className`, too.

kptdobe:
I do not like that. It adds a lot of css classes you might not want, looks like pollution. Let's go deeper and try to implement some use cases and see what is really useful and how to make use of this. See #198.

PR #138, opened 2018-12-14
[sections] missing types of nested elements #136

kptdobe:
PR for #136

This PR adds more logic on how we compute the "types":
- if the root is a paragraph, it analyses the types inside the 1st level children of that paragraph
- adds computing the number of occurrences for each type
 - improved unit testing


trieloff:
@kptdobe I know I told you at the Hackathon that I would not want to run this recursively, but for #112 it might make sense to run it recursively on lists.

trieloff:
Make sure to update the documentation here: https://github.com/adobe/helix-pipeline/blob/013081c4c61fdd210a0046f9bd1b39ca120c9db4/README.md#contentsections-in-detail – the `nb-*-*` class is not documented.

kptdobe:
> @kptdobe I know I told you at the Hackathon that I would not want to run this recursively, but for #112 it might make sense to run it recursively on lists.

This seems to make sense but it is a whole topic by itself. Let's tackle the "recursion for lists" as part of #112

PR #141, opened 2018-12-18
Consolidate request properties in dev and publish env #140

### LINKED ISSUE #140: Consolidate request properties in dev and publish env
There's currently a mismatch of request properties passed into the payload between dev and publish environments. And [`request.schema.json`](https://github.com/adobe/helix-pipeline/blob/master/docs/request.schema.json) says yet something different.

rofe:
Fixes #140

### DIFF
--- a/docs/request.schema.json
+++ b/docs/request.schema.json
@@ -52,10 +52,10 @@
             "type": "string",
             "description": "The HTTP method of the request",
             "meta:enum": {
-                "get": "GET request",
-                "post": "POST request",
-                "put": "PUT request (not handled by Helix)",
-                "delete": "DELETE request (not handled by Helix)"
+                "GET": "GET request",
+                "POST": "POST request",
+                "PUT": "PUT request (not handled by Helix)",
+                "DELETE": "DELETE request (not handled by Helix)"
             }
         },
         "headers": {
--- a/docs/request.schema.json
+++ b/docs/request.schema.json
@@ -29,7 +29,7 @@
             "type": "string",
             "description": "The path of the request URL",
             "examples": [
-                "/index.html"
+                "/index.md"
             ]
         },
         "selector": {
--- a/docs/request.schema.json
+++ b/docs/request.schema.json
@@ -50,7 +50,7 @@
         },
         "method": {
             "type": "string",
-            "description": "The HTTP method of the request",
+            "description": "The HTTP method of the request. Note: method names can be lower-case.",
             "meta:enum": {
                 "GET": "GET request",
                 "POST": "POST request",
--- a/docs/request.schema.json
+++ b/docs/request.schema.json
@@ -25,18 +25,11 @@
                 "/index.html?foo=bar"
             ]
         },
-        "resourcePath": {
-            "type": "string",
-            "description": "The resource path (without extension) that has been requested",
-            "examples": [
-                "/index"
-            ]
-        },
         "path": {
             "type": "string",
-            "description": "The path and request parameters of the request URL",
+            "description": "The path of the request URL",
             "examples": [
-                "/index.html?foo=bar"
+                "/index.html"
             ]
         },
         "selector": {
@@ -59,10 +52,10 @@
             "type": "string",
             "description": "The HTTP method of the request",
             "meta:enum": {
-                "GET": "GET request",
-                "POST": "POST request",
-                "PUT": "PUT request (not handled by Helix)",
-                "DELETE": "DELETE request (not handled by Helix)"
+                "get": "GET request",
+                "post": "POST request",
+                "put": "PUT request (not handled by Helix)",
+                "delete": "DELETE request (not handled by Helix)"
             }
         },
         "headers": {
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "@adobe/openwhisk-loggly-wrapper": "^0.3.0",
-    "ajv": "^6.5.4",
+    "ajv": "^6.6.2",
     "callsites": "^3.0.0",
     "clone": "^2.1.2",
     "fs-extra": "^7.0.0",
--- a/src/utils/openwhisk.js
+++ b/src/utils/openwhisk.js
@@ -20,9 +20,12 @@ const owwrapper = require('@adobe/openwhisk-loggly-wrapper');
  */
 function buildUrl(p) {
   let path = p.path || '/';
-  if (path.indexOf('.') !== -1) path = path.substring(0, path.lastIndexOf('.'));
+  const dot = path.lastIndexOf('.');
+  if (dot !== -1) {
+    path = path.substring(0, dot);
+  }
   const sel = p.selector ? `.${p.selector}` : '';
-  const ext = p.extension ? `.${p.extension}` : '';
+  const ext = `.${p.extension || 'html'}`;
   const query = p.params ? `?${p.params}` : '';
   return `${path}${sel}${ext}${query}`;
 }
--- a/src/utils/openwhisk.js
+++ b/src/utils/openwhisk.js
@@ -15,6 +15,19 @@ const querystring = require('querystring');
 const owwrapper = require('@adobe/openwhisk-loggly-wrapper');
 
 /**
+ * Builds the URL from path, selector, extension and params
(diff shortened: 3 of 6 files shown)


src/utils/openwhisk.js
```diff
@@ -33,6 +33,7 @@ function extractClientRequest(action) {
     path: request.params.path || '',
     extension: request.params.extension || '',
     selector: request.params.selector || '',
+    url: request.params.url || '',
```
tripodsan: as mentioned in https://github.com/adobe/helix-simulator/pull/112, this should be
`path + query` since the edge never sends the url.

package.json
```diff
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "@adobe/openwhisk-loggly-wrapper": "^0.3.0",
-    "ajv": "^6.5.4",
+    "ajv": "^6.6.2",
```
tripodsan: is this relevant?

docs/request.schema.json
```diff
@@ -25,18 +25,11 @@
                 "/index.html?foo=bar"
             ]
         },
-        "resourcePath": {
-            "type": "string",
-            "description": "The resource path (without extension) that has been requested",
-            "examples": [
-                "/index"
-            ]
-        },
```
tripodsan: why do you remove this? @trieloff, should we include it via fastly, compute it, or drop it completely? resource is a _old_ concept from sling, and we don't use it so far. :-)

docs/request.schema.json
```diff
@@ -59,10 +52,10 @@
             "type": "string",
             "description": "The HTTP method of the request",
             "meta:enum": {
-                "GET": "GET request",
-                "POST": "POST request",
-                "PUT": "PUT request (not handled by Helix)",
-                "DELETE": "DELETE request (not handled by Helix)"
+                "get": "GET request",
```
tripodsan: I think we should allow both: `get` and `GET`....

src/utils/openwhisk.js
rofe: now url is built on the fly

docs/request.schema.json
rofe: I just noticed that methods are always already lowercased, while the documentation had them uppercased...

@trieloff

package.json
rofe: no, just a result of rebuilding. should i remove it from this PR and commit it separately?

src/utils/openwhisk.js
```diff
@@ -14,6 +14,19 @@
 const querystring = require('querystring');
 const owwrapper = require('@adobe/openwhisk-loggly-wrapper');
 
+/**
+ * Builds the URL from path, selector, extension and params
+ * @param {Object} p The request parameters
+ */
+function buildUrl(p) {
+  let path = p.path || '/';
+  if (path.indexOf('.') !== -1) path = path.substring(0, path.lastIndexOf('.'));
```
tripodsan: no 1-liners for if. didn't eslint complain?

tripodsan: don't calculate lastIndexOf() twice. store it in a variable first.

docs/request.schema.json
tripodsan: what is the difference between `path` and `resourcePath+selector+extension` ?

test/testOpenwhisk.js
```diff
@@ -134,6 +134,7 @@ describe('Testing OpenWhisk adapter', () => {
         path: '/test',
         extension: 'html',
         selector: 'print.preview',
+        url: '/test.print.preview.html?a=42&b=green',
```
tripodsan: so...`path` is actually `resourcePath` ?

src/utils/openwhisk.js
tripodsan: also, I think the `path` here, is the path to the resource (i.e. the markdown file)

test/testOpenwhisk.js
tripodsan: also: add a test using a path to a md file, eg:

`/api/v1/webhtml?owner=tripodsan&repo=hlxtest&ref=master&path=/index.md&selector=&extension=html&strain=default&params=`

tripodsan:
@trieloff we probably should be more concise with the `path`. the one passed from the edge, should either be a `resourcePath` (w/o extension) or the original `clientRequestPath`. passing a _path_ to a markdown file is probably wrong, since the edge does not know about it.

package.json
trieloff: You can leave it in, it just might create a conflict if `master` is updated independently.

docs/request.schema.json
trieloff: I'd drop it. We can't do proper (read: Sling-like) resource resolution anyway and it would only be a duplication of the owner, repo, ref, path quadruple that we use to address content.

docs/request.schema.json
trieloff: I'd say we should revert OpenWhisk's lowercasing, but in order to do it consistently, we would also need to apply it to request headers, where it is not as reversible. Let's allow both.

trieloff: Some more context: the `meta:enum` property in the schema is for documentation, not validation. Therefore, I'd keep it uppercase in the `meta:enum`, but mention in the `description` that the method can be case insensitive. 

If we want to be more prescriptive with the methods allowed, we could add a `pattern`  or `enum` validation property, but that would rule out more creative HTTP method names,

src/utils/openwhisk.js
trieloff: I'd feel much better if we would use a URI parser here.

trieloff:
The parameters we pass from the edge are:

https://github.com/adobe/helix-cli/blob/385b3624c7ea162fe8e115ef9721e555e3a124d9/layouts/fastly/helix.vcl#L449-L457

* `owner`: obvious
* `repo`: obvious
* `ref`: obvious
* `path`: everything after the hostname or port and before the question mark (or end of the URL) **this is different from `path` in Node/Express, which includes the query string
* `extension`: everything from the last `.` in `path` to the end of `path`.
* `selector`: the substring of `path` that starts with `.` and ends at (does not include) the beginning of the extension

https://github.com/adobe/helix-cli/blob/385b3624c7ea162fe8e115ef9721e555e3a124d9/layouts/fastly/helix.vcl#L403-L406

src/utils/openwhisk.js
tripodsan: as @trieloff mentioned below, the path already incudes the selector/extension. so no need to construct it here.

docs/request.schema.json
rofe: ok

docs/request.schema.json
rofe: ok, reverted to uppercase and updated description.

test/testOpenwhisk.js
rofe: actually the test case is wrong: `path` always points to a md file. I fixed the test.

src/utils/openwhisk.js
rofe: > no 1-liners for if. didn't eslint complain?

no :) but I added a new block now

> don't calculate lastIndexOf() twice. store it in a variable first.

done

> the path already incudes the selector/extension. so no need to construct it here.

in my tests with [Adobe-Developer-Site](https://github.com/adobe/Adobe-Developer-Site), that isn't the case: `path` contains the path to the .md file, so we need to cut off the .md extension, add the `selector` and then `extension` from the `request.params`.

> I'd feel much better if we would use a URI parser here.

I can look into that next, however the current version behaves stably in my tests, with and without selector, even without extension. 

Examples:
```
{
  path: "/test.md",
  extension: "html",
  selector: "",
  query: "foo=bar"
}
```
returns `/test.html?foo=bar`, and
```
{
  path: "/test.md",
  selector: "dump",
  query: ""
}
```
returns `/test.dump.html`.

rofe: > I'd feel much better if we would use a URI parser here.

@trieloff I looked into it and don't think a parser would be much help in this case, because we have to compute the path segment of the `url` based on the md path, the selector (if it exists) and the extension. i therefore doubt that the code would become much more readable than it is now. But of course I wouldn't mind being convinced otherwise.

PR #163, opened 2019-01-18
Fix problems in the markdown -> html pipeline

koraa:
This fixes two issues with markdown rendering and adds an infrastructure for testing whether for a specific markdown input, the correct output is produced.

Not sure whether `test/testHTMLFromMarkdown.js` should not be merged into `test/testHTML.js`; I kind of like having all the markdown tests in their own place, on the other hand a lot of boilerplate is required to put them there…

### DIFF
--- a/src/utils/mdast-to-vdom.js
+++ b/src/utils/mdast-to-vdom.js
@@ -13,11 +13,10 @@
 /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^_" }] */
 const { selectAll } = require('unist-util-select');
 const handlers = require('mdast-util-to-hast/lib/handlers');
-const tohast = require('mdast-util-to-hast');
+const mdast2hast = require('mdast-util-to-hast');
+const hast2html = require('hast-util-to-html');
 const unified = require('unified');
 const parse = require('rehype-parse');
-const tohyper = require('hast-to-hyperscript');
-const h = require('hyperscript');
 const { JSDOM } = require('jsdom');
 const image = require('./image-handler');
 const embed = require('./embed-handler');
@@ -189,20 +188,13 @@ class VDOMTransformer {
   }
 
   /**
-   * Turns the MDAST into a basic DOM-like structure using Hyperscript
-   * @returns {Node} a simple DOM node (not all DOM functions exposed)
-   */
-  process() {
-    // turn MDAST to HTAST and then HTAST to VDOM via Hyperscript
-    return tohyper(h, tohast(this._root, { handlers: this._handlers }));
-  }
-
-  /**
    * Turns the MDAST into a full DOM-like structure using JSDOM
    * @returns {Node} a full DOM node
    */
   getDocument() {
-    return new JSDOM(this.process().outerHTML).window.document;
+    // mdast -> hast; hast -> html -> DOM using JSDOM
+    const hast = mdast2hast(this._root, { handlers: this._handlers });
+    return new JSDOM(hast2html(hast)).window.document;
   }
 }
 
--- a/docs/content.schema.json
+++ b/docs/content.schema.json
@@ -49,14 +49,6 @@
             "description": "Deprecated: the Hypermedia (HAST) AST",
             "meta:status": "deprecated"
         },
-        "children": {
-            "type": "array",
-            "description": "Deprecated: the main HTML elements of the document. `document.children[].innerHTML` instead.",
-            "items": {
-                "type": "string"
-            },
-            "meta:status": "deprecated"
-        },
         "html": {
             "type": "string",
             "meta:status": "deprecated",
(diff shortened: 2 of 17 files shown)


koraa:
I am also pondering what would be a good place for a domEquals function? Thinking of extracting one from assertDomEquals and putting those two functions somewhere where all helix code can use it…

trieloff:
We may want to create a new helix-testutils project. I need to put my `condit` method for conditional integration testing somewhere, too.

koraa:
After fixing the missing support for the missing "referenceType" attribute in the mdast json schema, I can't get the example below to work; needs extra investigation but I do not think this is any error we introduced, since this is a highly specific property of the GFM & CommonMark specs (which flavor are we using by the way).
Pondering whether this is worth fixing if it's somewhere deep inside mdast…in general IMHO it's worth fixing those really finicky errors because they make for even weirder edgecases and issues…

As far as I can see, the markdown is already being parsed incorrectly…it parses `[link]` as a reference and `(<foobar)` as adjacent text…

# Input Markdown

```markdown
# Foo

Hello World [link](<foobar)
```

# Expected result

(As rendered by marked with GFM enabled and without)

```html
<h1 id="foo">Foo</h1>
<p>Hello World <a href="%3Cfoobar">link</a></p>
```

# Actual Result

```html
<h1>Foo</h1>
<p>Hello World [link](&lt;foobar)</p>
```

# MDAST

```json
{ type: 'root',
  children:
   [ { type: 'heading',
       depth: 1,
       children:
        [ { type: 'text',
            value: 'Foo',
            position:
             Position {
               start: { line: 1, column: 3, offset: 2 },
               end: { line: 1, column: 6, offset: 5 },
               indent: [] } } ],
       position:
        Position {
          start: { line: 1, column: 1, offset: 0 },
          end: { line: 1, column: 6, offset: 5 },
          indent: [] } },
     { type: 'paragraph',
       children:
        [ { type: 'text',
            value: 'Hello World ',
            position:
             Position {
               start: { line: 3, column: 1, offset: 7 },
               end: { line: 3, column: 13, offset: 19 },
               indent: [] } },
          { type: 'linkReference',
            identifier: 'link',
            label: 'link',
            referenceType: 'shortcut',
            children:
             [ { type: 'text',
                 value: 'link',
                 position:
                  Position {
                    start: { line: 3, column: 14, offset: 20 },
                    end: { line: 3, column: 18, offset: 24 },
                    indent: [] } } ],
            position:
             Position {
               start: { line: 3, column: 13, offset: 19 },
               end: { line: 3, column: 19, offset: 25 },
               indent: [] } },
          { type: 'text',
            value: '(<foobar)',
            position:
             Position {
               start: { line: 3, column: 19, offset: 25 },
               end: { line: 3, column: 28, offset: 34 },
               indent: [] } } ],
       position:
        Position {
          start: { line: 3, column: 1, offset: 7 },
          end: { line: 3, column: 28, offset: 34 },
          indent: [] } } ],
  position:
   { start: { line: 1, column: 1, offset: 0 },
     end: { line: 3, column: 28, offset: 34 } } }
```

# HAST

```json
{ type: 'root',
  children:
   [ { type: 'element',
       tagName: 'h1',
       properties: {},
       children:
        [ { type: 'text',
            value: 'Foo',
            position:
             { start: { line: 1, column: 3, offset: 2 },
               end: { line: 1, column: 6, offset: 5 } } } ],
       position:
        { start: { line: 1, column: 1, offset: 0 },
          end: { line: 1, column: 6, offset: 5 } } },
     { type: 'text', value: '\n' },
     { type: 'element',
       tagName: 'p',
       properties: {},
       children:
        [ { type: 'text',
            value: 'Hello World ',
            position:
             { start: { line: 3, column: 1, offset: 7 },
               end: { line: 3, column: 13, offset: 19 } } },
          { type: 'text',
            value: '[link]',
            position:
             { start: { line: 3, column: 14, offset: 20 },
               end: { line: 3, column: 18, offset: 24 } } },
          { type: 'text',
            value: '(<foobar)',
            position:
             { start: { line: 3, column: 19, offset: 25 },
               end: { line: 3, column: 28, offset: 34 } } } ],
       position:
        { start: { line: 3, column: 1, offset: 7 },
          end: { line: 3, column: 28, offset: 34 } } } ],
  position:
   { start: { line: 1, column: 1, offset: 0 },
     end: { line: 3, column: 28, offset: 34 } } }
```

koraa:
Please do not merge yet; wip

src/utils/mdast-to-vdom.js
```diff
@@ -188,21 +187,14 @@ class VDOMTransformer {
     };
   }
 
-  /**
-   * Turns the MDAST into a basic DOM-like structure using Hyperscript
-   * @returns {Node} a simple DOM node (not all DOM functions exposed)
-   */
-  process() {
```
trieloff: The idea of this function is to be used in project-code, e.g. if you want to override the way HTML is being generated, you'd be able to register a new handler and then run `process` in your `pre.js`

trieloff: But I guess just using `getDocument` is easier.

kptdobe:
arff... no... too fast...
I use the `process` method because it returns a node only and not a full document! See https://github.com/kptdobe/helix-sections-playground/blob/master/src/html.pre.js#L50

PR #168, opened 2019-01-28
List Comprehension #112

kptdobe:
PR for #112


src/html/get-metadata.js
```diff
@@ -42,17 +42,41 @@ function image(section) {
...
+  }
+  return types;
+}
+
 /**
  * Sets the `types` attribute of the section, using following patterns:
  * 1. has-<type> for every type of content found in the section
  * 2. is-<type>-only for sections that have only content of type
  * 3. is-<type1>-<type2>-<type3> ranks the top three most common types of content
+ * 4. nb-<type>-<nb_occurences> is the number of occurences per typ
```
trieloff:
```suggestion
 * 4. nb-<type>-<nb_occurences> is the number of occurences per type
```

trieloff:
@kptdobe ready to merge this?

kptdobe:
Not sure, you wanted to move the childrenTypes to the mdast nodes. Did not change that yet.

kptdobe:
Revamped the PR, moved the "types" on the MDAST nodes. I used the `data` property which is described as the "place to be" for those can of extra info. See https://github.com/syntax-tree/unist#node

Sample usage can be seen here: https://github.com/kptdobe/helix-sections-playground/blob/list-comprehension-alignment/src/html.pre.js#L40-L49
Requesting http://localhost:3000/index.layout.html shows how the "types" for children can be added to each elements.

kptdobe:
Documented `data`. 
I also added properties supported by `mdast-util-to-hast`. See https://github.com/syntax-tree/mdast-util-to-hast/blob/master/readme.md#notes for the definition of `hName`, `hProperties` and `hChildren`. A full description of `hProperties` and `hChildren` is really overkill, we would have to document the DOM API and the HAST API...

`hProperties` is super useful to specify the css className for example at the MDAST level. The 2 others might be useful for really specific use cases. At least, the schema mention them.

PR #175, opened 2019-02-05
[WIP] [DISCUSSION] Sections Improvements

koraa:
I am currently investigating using sections in some examples in helix-cli. Upon investigation I found, that sections are currently  being generated; they feature one children property which contains mdast nodes and a 'type' node which is always set to root.

This strikes me as impractical, since users have to manually convert the mdast nodes to html themselves.
This [`pre.js`](https://github.com/adobe/developer.adobe.com/blob/master/src/default_html.pre.js) seems to confirm that the process of using sections is rather complex at the moment.

## Proposal

In order to make using sections easier for users, I propose that we redesign sections like this:

1. Add a free function named `sections(elm: element?)` for the htl to call.
2. Add a second function `sectionTitle(elm: element, default: any?)`.
3. Make sure the mdast -> dom/jsdom transformer emits a `<hr yamlData/>` for any yaml data in the document.

The `sections` function works like this:

1. If no `elm` is `undefined`, use `content.document.body`
2. If `elm` is a document node or document element, use `elm.body`
3. If `elm` is neither undefined, a document node/element or dom element, throw an error
4. Otherwise, the function generates a list of sections by iterating through the list of children in the given element:
  1. Each <section> tag is appended to the list of sections.
  2. All tags until an `<hr/>` or `<section>` element directly beneath `elm` are grouped into one section. These are wrapped into a virtual `<section>` element and appended to the list of sections.
  3. Implicit sections between a `<section>` tag and a `<section> or <hr/>` tag that are empty (containing only comments or textNodes containing white space) are discarded. Empty implicit sections between two `<hr/>` tags are RETAINED.
  4. All the sections in the list are wrapped into a `<body>` element. The sections can be inserted using `.innerHTML` or by iterating over `.children`

The `sectionTitle` functions works like this:

1. An appropriate dom node is resolved as in the first steps of the `sections` function.
2. The resulting element is recursively searched for any `<h*>` tag. 
  1. If a `<h*>` tag is found, we return it's innerHTML
  2. If no title is found and `default` is not `undefined`, we return `default`
  3. Otherwise we throw an exception

## Example

Using those two functions, we can easily construct a htl using sections:

```
<sly data-sly-list="${hlx.sections()}">
  <section title="${lodash.camelCase(hlx.sectionTitle(item))}">
    ${item.innerHTML}
  </section>
</sly>
```

### Generating virtual dom elements for  sections

An alternative approach would be to just yield a list of children instead, but generating a virtual section element provides the following advantages:

1. Looping through the sections' child nodes is possible `section.childNodes`, but not mandatory (`innerHTML` can be used)
2. Since no looping is required, there is no need to add a special case for textNodes (which do not have an `innerHTML` or `outerHTML`)

### Using free functions instead of pipeline steps

Instead of using free functions, we could also use the following strategies:

1. Move the section generation step to a point in the pipeline where the html has already been generated and generate a list of sections with titles there.`
2. Transforming each mdast node to html

Option two has the issue that further transformation dom transformations in the pipeline are ignored.
In both alternatives dom transformations in pre.js are ignored.

In my opinion, free functions are also more maintainable, scalable and facilitate deprecations more than extending our pipeline data structure; let me elaborate:

* Since using those functions is fully opt in, they are relatively cheap; pipeline steps in the default pipeline are always executed, even if their results are not used by the  actual htl, so they have to be very lightwight. Free functions invoked in the htl are pay-what-you-use. This also means we can add
a lot of utility functions if we like. A polimorphic `innerHTML()` that works on mdast, hast, document nodes, document elements and text nodes would be thinkable…
* The functions can be parameterized by the user; e.g. sectionTitle is made for sections, but could easily be used for other elements. For instance it could determine the title of an entire document.
* The call path to a complex transformation is made very obvious to the user; fewer surprises there!
* Since the functions are not used unless a user explicitly calls them, we can trivially add deprecation warnings
* We can also easily compose those functions in complex ways; this is not possible with pipeline functions


kptdobe:
I fully agree that sections as they are now are pretty unpractical.
The main idea behind sections was: the pipeline pre-computes the sections and assign to each of them a set of types (has-paragraph, np-image-2...). Those typed sections would go into a "layout magic machinery" that would determine for each section what kind of section it is (based on the types but also based on some end-user definition) and then determine how to render / style the section. We are not yet there for sure.
Then we introduced the VDOM API which introduced a lot of confusion because now we do not know which logic needs to happen on the MDAST or on the VDOM: styling should be on the DOM for sure but determining what is the meaning of a section makes more sense to happen on the MDAST.
And this should all happen somehow in the pre.js or in a declarative way.

If you look at the developer site, for now, they only need to style elements in the sections, i.e. add some CSS classes on the elements thus I agree that they should not have to worry about the MDAST vs VDOM or how to render the stuff. Any improvement in this direction is great!

But be careful with HTL spec as @raducotescu mentioned ;)

koraa:
@kptdobe Thanks for your thoughts!

To be honest, I am having trouble picturing some of this. Would you mind elaborating on some of your points?

Which part of the "layout magic machinery" and the section tagging should we implement for the start?

If we determine sections purely in mdast, how do we carry that info through the various transformation steps?
I mean currently AFAICS the sections are not subject to any of the transformations later in the html pipe? A situation might
arise where a customer has a lot of business logic working on their dom (`content.document`), wants to switch to
using sections but notices sections are incompatible with their existing pipeline...

kptdobe:
These are excellent questions and to be honest, I do not have an answer for them. The current state is early research phase I would say. Defining what are the next things to do is probably the top priority.

Before the last hackathon, I was in the process of re-building the https://www.tesla.com/model3 page with Helix. See https://github.com/kptdobe/helix-sections-playground/blob/master/t.md
There are 2 aspects of it:
1. as a content author, I need to focus on the content thus finding the simplest way to express in md all the content appearing on the page is a first step. My t.md is a really basic and simplified version of it.
2. as a developer, I need to build "something" that interpret whatever simple md an author would have output and make it look like my designer wants. 3 research parts in this:
  1. each section has its own constraints, the sections flow has some logic (vertifcal or horizontal list, banner above / below the image / video...). This flow must somehow be "scripted" or deterministic based on sections characteristic ("layout magic machinery" mentioned above). For now, this could be really simple and was not my top priority
  2. could a developer style the DOM that our pipeline outputs to look like the real Tesla page ?
  3. if not, thus how / where do we apply extra DOM transformation or define the desire outputs ? If you look at the DOM of the Tesla page and assume we would have to generate that, this would be hell to do in the current state. The developer site is in between where they do not need to change the DOM but simply apply css class names.

This is more than "sections" but they are the end result at the moment. How / when we render the sections is one piece of the problem only.

koraa:
I see! Seems like a very worthy end goal :)

For me, the styling stuff seems a bit large at the moment so I would prefer to put that off until some other time and just concentrate on getting the section splitting right…
Though I also think getting the sectioning right is a step in the direction you want to go! After all, sections by themselves are pretty much just div elements with a fancy name; the nice effects on the tesla website are all CSS :)

There probably are two basic issues to discuss:
* How can we reuse transformations on `content.document` late in the pipeline for sections.
* Do we want to use functions to let the user opt in to some heavy transformations

IMO for the first we migth just go the route of having a mdast pipeline step to annotate sections using special html tag. Or we might insert `---` (horizontal rule) into  the mdast since if we actually wanted to embed special html tags we would have to allow inline html in the mdast -> htast transformer.

kptdobe:
I am perfectly fine testing a prototype on a better way of splitting the sections ;)

tripodsan:
> free functions

what are free functions?

koraa:
@kptdobe Just a function that may be called by the htl

kptdobe:
You meant @tripodsan :)

tripodsan:
> Just a function that may be called by the htl

ok, but why is it called _free_ ?

koraa:
Technically speaking, a free function is any function that is not a member function…

Yeah, something like that! I thought maybe we could add another global `hlx` where we could just put arbitrary functions…
Making it a member of content does not seem right to me, unless we make content a class, since I would expect content do only contain the actual data, but that's just bike shedding :)

tripodsan:
> Technically speaking, a free function is any function that is not a member function…

this doesn't really match your example, where `hlx.sections()` returns the sections of the current context: 

`<sly data-sly-list="${hlx.sections()}">`

trieloff:
@koraa I like your alternative proposal (adding proper DOM nodes to each section) much better than the original:

- it is not tied to HTL, but works in any templating language
- it does not introduce a new pattern to working with pipeline payload

Furthermore, we should be clear if we are operating on an MDAST or a DOM. In my mind, the MDAST is always a better representation of the author's intent, so I'd prefer to stay as close to the MDAST as long as possible.

koraa:
@tripodsan Well, a free function with context is a relatively normal beast…especially in languages supporting imperative programming…

But yes, it is a bit odd. Unfortunately the alternative of providing a member method that does not use it's context at all if a parameter is given strikes me as considerable more odd…especially if the object we are attaching the method to is a plain object as opposed to the instance of a class…

Right now I am adding `sections()` and `findTitle()` to helix.shared anyways and I think this is probably the right place for them anyways…
So in my mind we could just do this:

1) Turn content() into a proper class and add `.sections()` as a method (not sure if this is anywhere close to where we want to develop the pipeline data in the long term. @trieloff what do you think)
2) If people want to customize sections and use it on another element, they can just manually import the `helix.shared.dom` module (is this even possible in htl? Imo would be nice if it was).

Could you (cc @kptdobe) please also quickly check out the doc I wrote for sections() and findTitle()? Unfortunately they turned out to be another cesspool of edge cases, so I am interested whether the handling matches your  mental models?

tripodsan:
> Well, a free function with context is a relatively normal beast…especially in languages supporting imperative programming…

all i'm saying is: as soon as a function is invoked where an argument is implied by the object it is invoked on (eg: `hlx.sections()`), i'd consider it as a method (and not a function anymore). I would even classify `section.call(hlx)` as a method call.

however, please also note, that htl does not support function calls in expressions directly. we would need to support use-classes. see https://github.com/adobe/htl-spec/blob/master/SPECIFICATION.md#42-javascript-use-api and https://sling.apache.org/documentation/bundles/scripting/scripting-htl.html

Once implemented, we could create a `sections` use class that we initialize with:

```
<sly data-sly-use.sections="${'@adobe/helix-pipeline:Sections' @ dom=content.document}" />
```

the `Sections` should be iterable, so we can do things like:

```
<sly data-sly-list="${sections}">
  <section title="${lodash.camelCase(hlx.sectionTitle(item))}">
    ${item.innerHTML}
  </section>
</sly>
```

koraa:
@tripodsan I see what you mean. I will finalize the sections() function in .shared and then we can deal with how we are going to integrate it here :)

trieloff:
> 1. Turn `content()` into a proper class and add `.sections()` as a method (not sure if this is anywhere close to where we want to develop the pipeline data in the long term. @trieloff what do you think)

Having proper objects in the payload (as opposed to the mostly dumb JSON we have right now) should be ok, as long as the object methods are pure and don't come with side effects or state changes.

koraa:
After a discussion with @trieloff some points where clarified; specifically some design considerations I hadn't until now considered emerged:

* Users should be encouraged to write their transformation steps in mdast instead of dom, because generally the dom has a lot of edge cases.
* One part of this is that it should be easy for users to register mdast transform steps; various proposals for specific implementations where discussed (generating the actual dom lazily from mdast, adding sections inside the mdast and providing a function that makes it easy to register mdast transform steps; simplifying the pipeline)
* Metadata extraction might be generally easier to do in mdast than in dom, so accessing sections as mdast might facilitate extracting metadata for each section

The last point is something that can be testes empirically; I will implement at least the following metadata extractors for mdast and dom each, so we can compare their complexity:

* section title
* section blurb

@trieloff Could you add the properties I missed please?
@trieloff I probably forgot some of the points you made today. Could you check if you remember some more and add these to the thread?

koraa:
**Columbo:** We also discussed using different templating engines/input formats. Once we have a working sections proposal, I will go ahead and test the sections proposal with some of these formats.
Which are the formats I should test?

@tripodsan I've been thinking and IMHO it would be great if users where just able to perform requires like in node inside their htl. `<require "package">` or maybe an actual js syntax `const lodash = require("lodash")`. IMO that would cover a lot of use cases where a user might be tempted to extend their pre.js…

tripodsan:
yes, I think something like:

```
<sly data-sly-use.sections="${'@adobe/helix-pipeline:Sections' @ dom=content.document}" />
```
 see https://github.com/adobe/htlengine/issues/42

koraa:
@tripodsan Ahh now I get it! Thanks for pointing out how sly works again.

Closed in reference to https://github.com/adobe/helix-pipeline/issues/185

Issue #155, opened 2019-01-09
md parser regression: setext headings no longer work


tripodsan:
markdown with headings written with underlines no longer work.

eg:

```
Main Title
=======

Sub Title
---------


````

render perfect in github:

Main Title
=======

Sub Title
---------

but are not detected anymore by the parser. eg:

```html
<title>Main Title
=======================</title>

<p>Sub Title</p>
```

rofe:
Seems like those are recognize as `paragraph` now (and the single underline as `thematicbreak`):
```
{
      "children": [
        {
          "type": "text",
          "value": "Main Title\n========="
        }
      ],
      "type": "paragraph"
    },
    {
      "children": [
        {
          "type": "text",
          "value": "Sub Title"
        }
      ],
      "type": "paragraph"
    },
    {
      "type": "thematicBreak"
    }
```

rofe:
Did this ever work? We updated `remark-parse` from 5.0.0 to 6.0.0 on Oct 23, and it seems that even v5 didn't parse these correctly...

tripodsan:
> Did this ever work?

yes. the initial examples in the demo was with `======`.

rofe:
Looks like setext headings were disabled deliberately in d45ab11a5a2f4191369cf4c8c2e21fde52e25fd7 to fix some grayscale issue: 
https://github.com/adobe/helix-pipeline/blame/master/src/html/parse-markdown.js#L43

@trieloff

trieloff:
I think the best solution would be to
1) fix the upstream https://github.com/remarkjs/remark-frontmatter so that it does not confuse setext headings with yaml fences
2) remove the lines that disable setext headings here

rofe:
For 1, it would be great if you could create the issue in remarkjs because you can best describe the original use case you worked around.
I can take care of 2.

PR #182, opened 2019-02-15
md parser regression: setext headings no longer work #155

rofe:
As discussed in #155, I removed the line im parse-markdown.js that caused setext headings to be ignored. Also added a test to show that both setext and # headings are processed correctly.

### DIFF
--- a/src/html/parse-markdown.js
+++ b/src/html/parse-markdown.js
@@ -40,7 +40,7 @@ function parse({ content: { body = '' } = {} }, { logger }) {
   logger.debug(`Parsing markdown from request body starting with ${body.split('\n')[0]}`);
 
   const preprocessor = unified()
-    .use(remark, { setext: false })
+    .use(remark)
     .use(frontmatter, { type: 'yaml', marker: '-', anywhere: true });
 
   // see https://github.com/syntax-tree/mdast for documentation
--- a/src/html/parse-markdown.js
+++ b/src/html/parse-markdown.js
@@ -39,9 +39,6 @@ const thbreak = {
 function parse({ content: { body = '' } = {} }, { logger }) {
   logger.debug(`Parsing markdown from request body starting with ${body.split('\n')[0]}`);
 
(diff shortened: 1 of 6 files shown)


rofe:
Fixes #155

src/html/parse-markdown.js
```diff
@@ -39,9 +39,6 @@ const thbreak = {
 function parse({ content: { body = '' } = {} }, { logger }) {
   logger.debug(`Parsing markdown from request body starting with ${body.split('\n')[0]}`);
 
-  // disable setext headings (=== and ---)
-  remark.Parser.prototype.blockTokenizers.setextHeading = () => {};
-
   const preprocessor = unified()
     .use(remark, { setext: false })
```
trieloff: Does this `{ setext: false }` flag do anything?

rofe: I was wondering the same thing actually. Our tests pass with and without this option, but I wasn't entirely certein and wanted to keep my changes as surgical as possible. `remark` [docs](https://github.com/remarkjs/remark/tree/master/packages/remark-stringify#optionssetext) suggest it does enable or disable setext headings, though...

rofe: That appears to be a `remark-stringify` option :) I think we can remove it.

rofe:
> Looks good. It would be great to test setext headlines in conjunction with YAML front matter and sections, too.

I tried replacing all level 1 and 2 atx headings in `frontmatter.md`(https://github.com/adobe/helix-pipeline/blob/master/test/fixtures/frontmatter.md) with setext format. Parsing works for all but [one](https://github.com/adobe/helix-pipeline/blame/master/test/fixtures/frontmatter.md#L144) which gets rendered as a `paragraph` followed by a `thematicBreak`...

rofe:
As discussed in [#155](https://github.com/adobe/helix-pipeline/issues/155#issuecomment-464837545) let's handle `frontmatter` in a separate upstream issue.

PR #188, opened 2019-02-21
display more useful output on errors #187

### LINKED ISSUE #187: display more useful output on errors
In case of an error somewhere in the `helix-pipeline`, the browser output is a bleak `""`. The developer will then have to look at the console output/log to find out what went wrong. 

Let's try to create a more meaningful response in case of an error.

(Partially implements https://github.com/adobe/helix-simulator/issues/17)

rofe:
Fixes #187

### DIFF
--- a/src/html/set-status.js
+++ b/src/html/set-status.js
@@ -22,7 +22,10 @@ function setStatus({ response = {}, error }, { logger }) {
     return {
       response: {
         status: 500,
-        body: '',
+        headers: {
+          'Content-Type': 'text/html',
+        },
+        body: `<html><body><h1>500</h1><p>${error}</p></body></html>`,
       },
     };
   }
--- a/src/html/set-status.js
+++ b/src/html/set-status.js
@@ -21,15 +21,18 @@ function setStatus({ response = {}, error }, { logger }) {
   if (error) {
     logger.debug('payload.error -> 500');
     const isDev = !production();
-    const res = {};
-    res.status = 500;
+    const res = {
+      status: 500,
+    };
     if (isDev) {
       res.body = `<html><body><h1>500</h1><p>${error}</p></body></html>`;
       res.headers = { 'Content-Type': 'text/html' };
     } else {
       res.body = '';
     }
-    return res;
+    return {
+      response: res,
+    };
   }
 
   return {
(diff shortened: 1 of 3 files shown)


trieloff:
See also https://github.com/adobe/helix-simulator/issues/17

rofe:
> Please check that the error output is only shown in non-production environments

Like so?

src/html/set-status.js
```diff
@@ -19,12 +20,16 @@ function setStatus({ response = {}, error }, { logger }) {
...
+    const isDev = !production();
+    const res = {};
+    res.status = 500;
+    if (isDev) {
+      res.body = `<html><body><h1>500</h1><p>${error}</p></body></html>`;
+      res.headers = { 'Content-Type': 'text/html' };
+    } else {
+      res.body = '';
+    }
+    return res;
```
tripodsan: you're missing a `response` object in your response :-)

```suggestion
    return {
       response: res,
    };
```

test/testSetStatus.js
```diff
@@ -20,16 +20,33 @@ const logger = Logger.getTestLogger({
...
+
   it('sets a 500 for an error', () => {
     assert.deepEqual(
-      setStatus({ content: { html: '<html></html>' }, error: 'oh, no!' }, { logger }),
+      setStatus({ content: { html: '<html></html>' }, error }, { logger }),
       {
-        response: {
-          status: 500,
-          body: '',
+        status: 500,
```
tripodsan: is this really correct? don't we need the `response` object?

src/html/set-status.js
rofe: Oops :)

test/testSetStatus.js
rofe: you're right.

rofe:
Readded lost `response` object (thanks @tripodsan!)

trieloff:
@rofe it's generally a better idea to pull the `if` statement out of the function and apply it in the pipeline like this:

```javascript
.error(detailed-status).when(() => !production())
.error(terse-status).when(() => production())
```

that way, your `detailed-status` and `terse-status` functions can stay simple, and you easily get 100% branch coverage in testing because there is only one branch in each function.

Issue #190, opened 2019-02-26
Adopt semantic-release


trieloff:
See https://github.com/adobe/developer.adobe.com/pull/144#issuecomment-467385381 – we should communicate breaking changes more clearly.

trieloff:
Closed in #192

PR #192, opened 2019-02-26
Introduce Semantic release

trieloff:
Fix for #190 

Biggest change for developers is described in the update to `CONTRIBUTING.md`: you have to run `npm run commit` to generate a commit message. Upon merging into `master`, the release will automatically get a semantically appropriate version number.


trieloff:
> Can we add a commit hook, to check if the proper commit format is used?

I don't know if this would play well with Greenkeeper and GitHub's squashes, but we can try later on.

Issue #167, opened 2019-01-28
Bring back process


trieloff:
> arff... no... too fast...
> I use the `process` method because it returns a node only and not a full document! See > https://github.com/kptdobe/helix-sections-playground/blob/master/src/html.pre.js#L50

_Originally posted by @kptdobe in https://github.com/adobe/helix-pipeline/pull/163#issuecomment-457817570_

kptdobe:
Thanks @trieloff 

We could just expose a `getNode` method instead of `process`. You would then only have `getDocument` or `getNode`.

PR #194, opened 2019-03-04
Bring back process

kptdobe:
As defined in #167, introduced the `VDOMTransformer#getNode` method (in lieu of `VDOMTransformer#process`): returns a node only and not a full document. This allows to do things like: 

```js
const transformer = new VDOM(sectionMDAST, action.secrets);
const node = transformer.getNode('section');
node.classList.add('mycssclass');
console.log(node.outerHTML);
```

This would output something like:

```html
<section class="mycssclass">
   <p><img src="/content/images/banner1.jpg" alt="banner1"></p>
   <h1>Helix - sections</h1>
   <p>Sample project to illustrate how sections work</p>
</section>
```

Same could be achieved with `VDOMTransformer#getDocument` but requires some transformations (wrapper in that case is the `body` element which makes things a bit more complicated). 

The returned Node element even has methods like `getElementsByTagName` or `getElementsByClassName` which are normally `document` methods!


koraa:
Can you give an example of a use case?
My immediate reaction was that we might als be able to solve this using .innerHTML on body like this in htl:

```
<div class="mycustomwrapper">
  ${document.innerHTML}
</div>
```

kptdobe:
In the HTL, yes. Well, almost. It would be: 

```html
<div class="mycustomwrapper">
  ${document.body.innerHTML}
</div>
```
(assuming `document = transformer.getDocument()`).
I personally do not like this `body` here (there is also a `head` actually since this is a complete `document`). This is used in the context of the sections which are only fragments.

What if you want to do the same in your `pre.js` ? When working with `sections`, my use case is always the same: I define the tag (like `<section>`) and compute the css class names of the container. The proposed `getNode` gives me "the fragment" and I do not need to care about the `body`. That's basically a shortcut method since this is all doable with the `getDocument` method.

koraa:
we could add a getBody() that just returns `getDocument().body`?

koraa:
I mean the main thing I would like to address is all the code duplication from getDocument; but in the end getNode could be just implemented as a wrapper around getDocument.

kptdobe:
Good feedback! I improved the implementation to reduce the code duplication. Maybe it is more clear now (still working on the tests of the new static function). 
The `getBody` is not a good option neither, I do not want to have a body!

The more I work with the VDOMTransformer, the more I am convinced there are 2 cases. If you have an MAST like (schematic):

```json
{
  root
  children: [
    {
      image
      url: xyz
    },
    {
      paragraph
      ...
    },
  ]
}
```

`getDocument`should give you an object that contains:

```html
<html>
  <head></head>
  <body>
    <img src ="xyz">
    <p>...</p>
  </body>
</html>
```

`getNode`should give you an object that contains:

```html
<div>
  <img src ="xyz">
  <p>...</p>
</div>
```

Like this the transformer allows you to quickly get either a full document or a snippet / fragment.

kptdobe:
arff.. just realized we have no test for all the static methods of VDOMTransformer...

kptdobe:
Added the tests for the `toHTML` method and created https://github.com/adobe/helix-pipeline/issues/195 for the missing tests.

PR #199, opened 2019-03-06
Rename is-<type>-only to has-only-<type>

kptdobe:
One first PR for https://github.com/adobe/helix-pipeline/issues/198

2 remaining questions for @trieloff: 

1. Does this rename have an impact on the TypeMatcher ? I do not think so but the impl is obscur, I am not sure.
1. One test (https://github.com/adobe/helix-pipeline/blob/master/test/testConditionalSections.js#L432) is failing but I think it was working by accident. I spent the day debugging this test and concluded that the test is wrong and maybe illustrating a wrong implementation / expectation. The code that selects one section or another is https://github.com/adobe/helix-pipeline/blob/master/src/utils/conditional-sections.js#L92 and changing the line https://github.com/adobe/helix-pipeline/blob/master/src/html/get-metadata.js#L53 with `types.push(`is-${Object.keys(typecounter)[0]}-only123`);` also breaks the test. Those 2 returned objects are different as written in your test but, this is a coincidence, not something predictable. Any thoughts ? Maybe I missed something but I cannot see how this could work.


trieloff:
1: should not impact the TypeMatcher
2: conditional sections for A/B testing use the strain name and the hash of the section content to randomly, but stably select one alternative. When you change the section's MDAST content, it changes the hash, with a 50% chance new hash is smaller than the alternative sections' hash, which means it gets taken and the test fails.. The fix is to wiggle either the strain name or the section content a bit until randomness favors us. See 46235fe

trieloff:
@kptdobe make sure you commit using `npm run commit`, so that we get a new release/version.

kptdobe:
> conditional sections for A/B testing use the strain name and the hash of the section content to randomly, but stably select one alternative. When you change the section's MDAST content, it changes the hash, with a 50% chance new hash is smaller than the alternative sections' hash, which means it gets taken and the test fails.. The fix is to wiggle either the strain name or the section content a bit until randomness favors us. See 46235fe

I am not sure I fully understand this in the context of someone trying to use the feature! How can you do A/B testing if in certain circumstances you end up having a final hash that always leads to the same decision...? There is a probability that all the hash + strain name leads to the same section. Thus the A/B testing results are not relevant. 
Why not using a random number instead ? You would have a "real" 50% chance in that case. (which I agree would be hard to test!)

trieloff:
The idea is that you'd randomly assign visitors to a strain (something we don't have yet), but once you are in a strain, every one in that strain gets the same content. `hash(content + strain)` is effectively a random number, but it does not change on every request.

If we could find something that would create more stable hashes of content, I'd take it.

kptdobe:
Humm... if you randomly assign visitors to a strain, why not letting the author decide and specify the test and the strain. 
```md
---
frontmatter: true
---

This is an easy test.

***

These two sections should always be shown

---
test: a
strain: randomGroup1
---

But neither this one.

---
test: a
strain: randomGroup2
---

Or this one at the same time.

```

The `strain` could even be a regex if you have multiple groups.

But I think we should get more clarity on how this would be set up and used. Maybe illustrate with a real life example and poc.

trieloff:
You can already force pin a section to a strain (using `strain: returning-visitors` for instance). 

With `test` you can create consistent test scenarios across multiple pages of the site and you don't have to do anything to balance the distribution and it works even when new strains are introduced.

kptdobe:
For now we only have one strain for the main publish (simple scenario, let's call it `prod`). If you want to set this up, you might create 1 new strain that fully inherits from `prod`, let's call it `prod-4abtest`. Assuming you have some magic somewhere that sets 50% of incoming users into `prod` and 50% in `prod-4abtest`, the only solution for a proper A/B testing is that the `hash(content + 'prod')` is smaller  than the `hash(content + 'prod-4abtest')` which is something we cannot guaranty. In 50% of the cases, the setup is broken and understanding why requires a lot of knowledge - not to say is possible only for you.

trieloff:
Hm. I'm pretty sure there is no solution that forces a different selection for any pair of two strains. When you have two variants (1 and 2) and three strains (A, B, C), you won't be able to enforce each strain to be different.

As the strains cannot know anything about the content and the page content cannot know anything about content on other pages or what strains exist, I believe a somewhat stable random assignment is the best thing to shoot for.

PR #206, opened 2019-03-14
Refactor frontmatter support

koraa:
feat(html pipe): Manual & improved frontmatter parsing

* Reduce number of dependencies
* Yaml is accessible in the main mdast, not only in section metadata
* Parsing & decoding frontmatter in one step
* Explicit constraints on what is parsed as frontmatter and what is not
  Before this we  would interpret pretty much anything as a frontmatter block and juust discard blocks that happened to not be valid yaml. This PR decouples the frontmatter detection from the actual content parsing.
* Helpful error messages for ambigous frontmatter blocks & frontmatter
  blocks that could not be parsed
* Reduces complexity in how yaml constraints are applied: No need to
  revert frontmatter blocks to normal markdown when constraints
  disqualify them; these blocks are just never converted to yaml ast
  nodes
* Handles a few edge cases better

```
Paragraph.

---

Item One: Clearly two description style paragraphs sandwhiched between <hr>

Item Two: Which are instead interpreted as yaml!

---
```

```
Paragraph.

---

Heading with colon: Right after <hr>
---
```

```
Paragraph.

---

+ Description style: list with plus
+ That is: sandwiched between <hr>

---
```

### DIFF
--- a/src/defaults/html.pipe.js
+++ b/src/defaults/html.pipe.js
@@ -30,6 +30,7 @@ const dump = require('../utils/dump-context.js');
 const validate = require('../utils/validate');
 const { cache, uncached } = require('../html/shared-cache');
 const embeds = require('../html/find-embeds');
+const parseFrontmatter = require('../html/parse-frontmatter');
 
 /* eslint no-param-reassign: off */
 /* eslint newline-per-chained-call: off */
@@ -47,6 +48,7 @@ const htmlpipe = (cont, payload, action) => {
     .every(validate).when(() => !production())
     .before(fetch).expose('fetch').when(hascontent)
     .before(parse).expose('parse')
+    .before(parseFrontmatter)
     .before(embeds)
     .before(smartypants)
     .before(sections)
--- a/src/html/parse-frontmatter.js
+++ b/src/html/parse-frontmatter.js
@@ -1,0 +1,307 @@
+/*
+ * Copyright 2018 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+const yaml = require('js-yaml');
+const { cloneDeep } = require('lodash');
+const {
+  join, map, zipLeast2, pipe, list, lookahead, filter, enumerate,
+  identity, range, size, type, reject,
+} = require('@adobe/helix-shared').sequence;
+
+const { assign } = Object;
+
+/**
+ * Given an mdast tree and it's text representation, this finds all
+ * the frontmatter in it.
+ *
+ * Frontmatter looks like this:
+ *
+ * ```
+ * ---
+ * Foo: bar
+ * ---
+ * ```
+ *
+ * The frontmatter is delimited by a `thematicBreak` made from three
+ * dashes in the mdast and contains a yaml-encoded payload.
+ *
+ * Not any yaml and any thematicBreak is accepted; in order to avoid
+ * false positives, the following restrictions apply:
+ *
+ * - There must be the start or end of the document or an empty line (`\n\n`)
+ *   before and after the frontmatter block
+ * - There may be no empty line within the frontmatter block; not even lines
+ *   containing only whitespace
+ * - The yaml must yield an object (as in key-value pairs); strings, numbers or
+ *   arrays are not considered to be frontmatter in this context
+ * - The thematic break must be made up of three dashes
+ *
+ * Note that most of the information required to asses these properties
+ * is not contained in the mdast itself, which is why this algorithm requires
+ * access to the original markdown string. (The mdast is an Abstract Syntax Tree,
+ * the proper tool for a task like this would be a Concrete Syntax Tree, but we have
+ * no such thing...).
+ *
+ * Note that converting the mdast to a markdown string will not do, since
+ * the generated markdown will be much different.
+ *
+ * # Future directions
+ *
+ * This function is likely to change in the following ways:
+ *
+ * - Relaxed restrictions on what constitutes frontmatter.
+ * - The ability to specify custom formats; by specifying the name
+ *   of the format explicitly we reduce ambiguity to pretty much zero
+ *   and can allow for far more complex frontmatter formats.
+ *
+ *    ```
+ *    ---json
+ *    {"foo": 42}
+ *    ---
+ *    ```
+ *
+ * # Ambiguous Frontmatter
+ *
+ * Normally, when one of the conditions above is triggered,
+ * (e.g. Frontmatter containing an empty line; not being an array
+ * instead of an object), a warning will be emitted instead of the
+ * frontmatter being actually parsed & returned.
+ * This warning may be processed by the caller in any way; e.g. by
+ * printing a warning on the console or by throwing an error...
+ *
+ * In order to avoid ambiguous cases, the format described above
+ * should be used for valid frontmatter; in order to use horizontal
+ * rules unambiguously, the markdown author should either use symbols
+ * other than dash to mark horizontal rules, or leave at least one empty
+ * line before and after the three dashes.
+ *
+ * Both ways are guaranteed to be interpreted as horizontal rules and
+ * never yield warnings.
+ *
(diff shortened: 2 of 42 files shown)


src/html/parse-frontmatter.js
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * The frontmatter is delimited by a `thematicBreak` made from three
+ * dashes in the mdast and contains a yaml-encoded payload.
+ *
+ * Not any yaml and any thematicBreak is accepted; in order to avoid
+ * false positives, the following restrictions apply:
+ *
+ * - There must be the start or end of the document or an empty line (`\n\n`)
+ *   before and after the frontmatter block
+ * - There may be no empty line within the frontmatter block; not even lines
```
trieloff: That might be too strict. Can we losen "within" to "at the start or end"?

src/html/parse-frontmatter.js
```diff
@@ -0,0 +1,206 @@
...
+    }),
+    // Group the fences into actual pairs of start/end fence
+    slidingWindow(2),
+    // Get rid of pseudo frontmatter blocks that sliding window
+    // implicitly inserts between the end of an actual frontmatter
+    // block and the start of the next
+    filter(([fst, last]) => fst.after && last.before),
+  );
+
+  const warn = (fst, last, txt, cause, prosa) => ({
```
trieloff: I'd use the Winston logger we pass as part of `action.logger`.

trieloff:
This pull request **introduces 2 alerts** when merging 6a67a7ca5462a8b7d7a78973ad235765cde72780 into 3e320e66e352853f49daf95c50529c5d59844e07 - [view on LGTM.com](https://lgtm.com/projects/g/adobe/helix-pipeline/rev/pr-bcaac6f0a29121ce4aedff8c26646b964df93b3b)

**new alerts:**

* 1 for Syntax error
* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

trieloff:
This pull request **introduces 1 alert** when merging cd2fd2b19d2234bdd60e979e435a94836654d81e into 6ddf1690518665e4056e5f83899c7c639c852071 - [view on LGTM.com](https://lgtm.com/projects/g/adobe/helix-pipeline/rev/pr-dc45db46350032bbbb441f10b4b7b9560fdae108)

**new alerts:**

* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

trieloff:
This pull request **introduces 1 alert** when merging 0846a48b5fdc9ef9a8761f1176f7b83a7511e5e0 into 6ddf1690518665e4056e5f83899c7c639c852071 - [view on LGTM.com](https://lgtm.com/projects/g/adobe/helix-pipeline/rev/pr-ae4a5a13c15fd54e4a2e393bd329fdcadbfd0fb7)

**new alerts:**

* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

trieloff:
This pull request **introduces 1 alert** when merging 336b5bc18046bfbbaa3e057aae16767946fdb7eb into 6ddf1690518665e4056e5f83899c7c639c852071 - [view on LGTM.com](https://lgtm.com/projects/g/adobe/helix-pipeline/rev/pr-7a3912be373aeb8df5bb386b8c5c0cc4fb19d8ca)

**new alerts:**

* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

trieloff:
This pull request **introduces 1 alert** when merging 6f4df45575823895314ffeefe7455f0f7f3c1177 into 6ddf1690518665e4056e5f83899c7c639c852071 - [view on LGTM.com](https://lgtm.com/projects/g/adobe/helix-pipeline/rev/pr-b65d898349b3414f4d81d4ea472ce0d255eb37d6)

**new alerts:**

* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

koraa:
Ok, this now finally, but it turns out it's complex enough to warrant a bit of refactoring…

src/html/parse-frontmatter.js
koraa: It's mean as an API for the caller. The caller can decide hot to process the warning…
The frontmatter pipeline step actually throws those warnings as errors ATM

src/html/parse-frontmatter.js
koraa: I've considered over the last couple days, and tbh I don't really know if this can be done safely…
We can introduce explicit frontmatter `---yaml` which should make false positives impossible and then also reintroduce empty lines in frontmatter…

koraa:
@trieloff I've run into some problems last friday when trying to add the frontmatter parse step into the actual pipline and this morning I was finally able to figure out the problem…
The pipeline property merger is adding all the mdast nodes to the mdast which I previously replaced with frontmatter nodes.
How can we disable that behaviour?

Btw, I think we might want to think about how we approach property merging in general again…
To me it seems that merging arrays is generally not what we want; AFAICS using _.merge also means we cannot delete data in pipeline steps, as the merging step will just reverse that deletion. Merging complex ASTs like mdast and htast could especially lead to weirdness deep in the tree…

I think the last two points might be espeically confusing to third party devs trying to write their own pipeline steps…

src/defaults/html.pipe.js
```diff
@@ -43,8 +44,11 @@ const htmlpipe = (cont, payload, action) => {
     .before(fetch)
     .when(({ content }) => !(content !== undefined && content.body !== undefined))
     .before(parse)
+    .before(parse_frontmatter)
+    .before(({ content }) => console.log("AFT PARSE_FM", content.mdast))
```
trieloff: In development mode, we dump the entire pipeline content into the `logs/debug` directory.

koraa: Jaja, das ist auch nur für mich zum debuggen…

koraa:
Ok, yay, we're down to 3 failing tests! So I had another interesting debugging experience yesterday; test/testConditionalSections.js:434 was failing and it took my a while to figure out why. Finally at some point I was able  to track down that the sorting order in strainhashsort had changed because the mdast format is now slightly different (payload containing the decoded yaml instead of value containing the yaml as a string).
The test was failing because the seed/strain chosen happened to yield a different ordering with the old  format than with the new. I was able to fix the test by trying other seeds until one worked.

So in the end the actual bug was in the tests, not in the code…

There still is a bit of an issue, because any changes in the section format or the mdast format might trigger a new test failuer; I could of course add a comment, but I am wondering whether we could just change the test selector to use an algorithm like `floor(number_of_options * hash(url + test + strain) / 2**hash_length)` (I avoid the typical modulo selection due to [modulo bias](https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator) unless I have a reson not to)?

This algorithm should have the same properties (yielding uncorrelated indices for different urls, strains and tests) while not depending on the actual ast or section format? It should even be a bit faster, because we do not have to process the entire ast?

trieloff:
This pull request **introduces 1 alert** when merging 8a4e384d17be1c03ecf29e8b875bb1144db0ef47 into b3b72293beb458aaf24353669b3b76130786cca5 - [view on LGTM.com](https://lgtm.com/projects/g/adobe/helix-pipeline/rev/pr-9ce3926d5b3d6f35d1c2d8867028d824cb79bc72)

**new alerts:**

* 1 for Unused variable, import, function or class

---

*Comment posted by [LGTM.com](https://lgtm.com)*

trieloff:
Regarding `testConditionalSections.js`, @kptdobe and I had a discussion a couple of days ago: https://github.com/adobe/helix-pipeline/pull/199#issuecomment-470448065

I think your suggestion does not look at the contents of the section at all, which means that either all sections in a test group get included or none.

An alternative I haven't considered in the discussion with Alex is to simply use the index of the section instead of the content hash. The index should be relatively stable.

koraa:
Not sure I understand your comment? 

> I think your suggestion does not look at the contents of the section at all, which means that either all sections in a test group get included or none.

> An alternative I haven't considered in the discussion with Alex is to simply use the index of the section instead of the content hash. The index should be relatively stable.

Hu? I was suggesting a way to determine that index: By using a hash over strain, test and url; since this would determine the index, it would always choose exactly one of the test selections, never more or less than that.

It would also look at the section.meta.test property, but not at the mdast or any other sections properties.
Am I missing something here?

trieloff:
The `test` property denotes the test group. Out of one test group, only one alternative gets chosen, so you might have a document that looks like this:

```
- section (test: intro)
- section (test: intro)
- section
- section (test: end)
- section (test: end)
```

The challenge is to select exactly one section for the intro and one for the end. By only including URL, Strain, and `test`, there is no difference between the two `intro` sections and no difference between the two `end` sections, which means either both or none would get chosen. We have to include at least one thing that is specific to the section itself.

Issue #159, opened 2019-01-14
Link references are crashing the pipeline


koraa:
```
# Foo

Hello -World-

[foo]
```

```
# Foo

Hello World [foo][bar]

[bar]: /url "title"
```

Possibly related #156

trieloff:
I'm seeing a similar issue with this Markdown:

```markdown
> [!NOTE]
> This is a note. Who'd have noted?
```

creating following MDAST:

```json
        {
          "type": "blockquote",
          "children": [
            {
              "type": "paragraph",
              "children": [
                {
                  "type": "linkReference",
                  "identifier": "!note",
                  "label": "!NOTE",
                  "referenceType": "shortcut",
                  "children": [
                    {
                      "type": "text",
                      "value": "!NOTE",
                      "position": {
                        "start": {
                          "line": 9,
                          "column": 4,
                          "offset": 90
                        },
                        "end": {
                          "line": 9,
                          "column": 9,
                          "offset": 95
                        },
                        "indent": []
                      }
                    }
                  ],
                  "position": {
                    "start": {
                      "line": 9,
                      "column": 3,
                      "offset": 89
                    },
                    "end": {
                      "line": 9,
                      "column": 10,
                      "offset": 96
                    },
                    "indent": []
                  }
                },
                {
                  "type": "text",
                  "value": "\nThis is a note. Who'd have noted?",
                  "position": {
                    "start": {
                      "line": 9,
                      "column": 10,
                      "offset": 96
                    },
                    "end": {
                      "line": 10,
                      "column": 36,
                      "offset": 132
                    },
                    "indent": [
                      3
                    ]
                  }
                }
              ],
              "position": {
                "start": {
                  "line": 9,
                  "column": 3,
                  "offset": 89
                },
                "end": {
                  "line": 10,
                  "column": 36,
                  "offset": 132
                },
                "indent": [
                  3
                ]
              }
            }
          ],
          "position": {
            "start": {
              "line": 9,
              "column": 1,
              "offset": 87
            },
            "end": {
              "line": 10,
              "column": 36,
              "offset": 132
            },
            "indent": [
              1
            ]
          }
        }
```

and following error message: 

```
[hlx] warn: Invalid Context at step 3, data.content.mdast.children[3].children[0].children[0] should NOT have additional properties
[hlx] warn: tapping failed: Error: Invalid Context at step 3
data.content.mdast.children[3].children[0].children[0] should NOT have additional propertiesInvalid Context at step 3
[...]

PR #211, opened 2019-03-15
Advanced extensibility support 🍝

trieloff:
One problem with our current pre.js model is that it can lead to maccaroni code
- just like spaghetti code, it can get messy
- but it's made of pipelines
- and you try to squeeze a lot of functionality through a single extension point

With this model, we can open multiple extension points along the pipeline, so that
a developer can inject custom functions before or after a particular pipeline step,
leading to better-stuctured penne code.

The attaching works through the `Pipeline.attach.before` and `Pipeline.attach.after`
functions which take each a pair of a function (like your pre function) and the name
of a pipeline step.

To make things easier, I plan to extend the wrapper in `helix-cli`, so that it carries
over `module.exports.before` and `module.exports.after` in a `pre.js`, just like
`module.exports.pre`. This would enable easy declarative extensibility without having
to override the entire pipeline.


src/defaults/html.pipe.js
```diff
@@ -61,6 +61,10 @@ const htmlpipe = (cont, payload, action) => {
     .when(esi) // flag ESI when there is ESI in the response
     .error(selectStatus(production()));
 
+
+  // pipe.attach.before(foo, 'fetch');
```
koraa: I suggest taking the function as a second argument…much more ergonomic…

src/pipeline.js
```diff
@@ -94,6 +94,37 @@ class Pipeline {
     this._posts = [];
     // functions that are executed before each step
     this._taps = [];
+
+    this.attach = function attach(ext) {
+      if (this.sealed) {
+        return;
```
koraa: I'd throw an error instead, as this might indicate a developer error…

src/pipeline.js
```diff
@@ -94,6 +94,37 @@ class Pipeline {
...
+      const foundposts = this._posts
+        .filter(post => post.alias)
+        .findIndex(post => re.test(post.alias));
+      if (foundpres !== -1) {
+        this._pres.splice(foundpres + offset, 0, f);
+      } else if (foundposts !== -1) {
+        this._posts.splice(foundpres + offset, 0, f);
+      }
+    };
+    this.attach.before = (f, name) => this.attach.generic(f, name, 0);
```
koraa: Could you please write some JSDOC?

src/pipeline.js
```diff
@@ -94,6 +94,37 @@ class Pipeline {
...
+        Object.keys(ext.before).map(key => this.before(ext.before[key], key));
+      }
+      if (ext && ext.after && typeof ext.after === 'object') {
+        Object.keys(ext.after).map(key => this.after(ext.after[key], key));
+      }
+      this.sealed = true;
+    };
+
+    this.attach.generic = (f, name, offset = 0) => {
+      const re = new RegExp(`^.*\\:${name} from `);
```
koraa: I am afraid I am having a bit of trouble understanding this function. Could you add some comments to elaborate on how the implementation works?

src/defaults/html.pipe.js
koraa: Oh and are these supposed to be comments?

koraa:
Oh, and I think you might forgot adding tests for the attach function itself? Or am I missing something here?

trieloff:
@koraa added tests in 63f9a6c

koraa:
thx!

koraa:
Could we also do a test extending the actual html pipeline? Maybe just extend some of the steps like `fetch` or the try injecting something at the end of the mdast part of the pipeline?

src/defaults/html.pipe.js
trieloff: Removed them.

src/pipeline.js
trieloff: No, running the pipeline twice is ok. I just don't want to re-add the extension points.

trieloff:
@koraa see 89ea8f5 for a slightly more realistic HTML pipeline extension. I'd build the real deal as part of the integration test here: https://github.com/adobe/helix-cli/pull/653

src/pipeline.js
```diff
@@ -94,6 +94,67 @@ class Pipeline {
...
+        Object.keys(ext.after).map(key => this.attach.after(key, ext.after[key]));
+      }
+      this.sealed = true;
+    };
+
+    /**
+     * Registers an extension to the pipeline.
+     * @param {String} name - name of the extension point (typically the function name).
+     * @param {pipelineFunction} f - a new pipeline step that will be injected relative to `name`.
+     * @param {integer} offset - where to insert the new function (0 = before, 1 = after)
```
tripodsan: can we have constants for the offset? or do you plan to add real numerical offsets?
why not a boolean?
why not `-1` for _before_?

src/pipeline.js
tripodsan: how does a developer _know_ the name of the functions?
after all, it is very volatile and defined by the `pipe` function used.

src/pipeline.js
```diff
@@ -94,6 +94,67 @@ class Pipeline {
...
+     */
+    this.attach.before = (name, f) => this.attach.generic(name, f, 0);
+    /**
+     * Registers an extension to the pipeline. The function `f` will be run in
+     * the pipeline after the function called `name` will be executed. If `name`
+     * does not exist, `f` will never be executed.
+     * @param {String} name - name of the extension point (typically the function name).
+     * @param {pipelineFunction} f - a new pipeline step that will be injected relative to `name`.
+     */
+    this.attach.after = (name, f) => this.attach.generic(name, f, 2);
```
tripodsan: why `2`? above you write `1 = after`

src/pipeline.js
```diff
@@ -112,9 +112,10 @@ class Pipeline {
      * Registers an extension to the pipeline.
      * @param {String} name - name of the extension point (typically the function name).
      * @param {pipelineFunction} f - a new pipeline step that will be injected relative to `name`.
-     * @param {integer} offset - where to insert the new function (0 = before, 1 = after)
+     * @param {integer} before - where to insert the new function (true = before, false = after)
      */
-    this.attach.generic = (name, f, offset = 0) => {
+    this.attach.generic = (name, f, before = true) => {
```
tripodsan: pedantic comment: good practice is to have `false` as default...but since the invention of defaults for arguments, this is no longer so important :-)

tripodsan: btw: do you really want to export `attach.generic` ?

src/pipeline.js
```diff
@@ -146,15 +147,15 @@ class Pipeline {
      * @param {String} name - name of the extension point (typically the function name).
      * @param {pipelineFunction} f - a new pipeline step that will be injected relative to `name`.
      */
-    this.attach.before = (name, f) => this.attach.generic(name, f, 0);
+    this.attach.before = (name, f) => this.attach.generic(name, f, true);
```
tripodsan: if `attachGeneric(before, name, f)` is a function, you could write:

```
this.attach.before = attachGeneric.bind(this, true);
```

trieloff:
@tripodsan you've confirmed the one concern I had with this approach: that we would allow to grow the API surface uncontrollably. With 461d3f6 all extension points are named and made explicit by the pipeline author.

src/pipeline.js
trieloff: No.

src/pipeline.js
trieloff: I used a similar approach in 461d3f6

trieloff:
@koraa explicit names are now the only way to use this. The pipeline author can set the explicit name for each function, but in order to keep the API stable, in a separate call like this:

```javascript
pipe.before(foo).ext('fooizer')
```

rather than

```
`pipe.before(foo, 'fooizer')`
```

Of course, you can set the property explicitly in the function definition, but I think the definition in the pipeline makes for more readable code, because you can immediately see where the extension points are.

trieloff:
I don't have a good solution for the order issue you've mentioned other than making `extend.after` and `extend.before` take lists of functions. Given that the most common usage will be through the shorthand syntax in `pre.js`, I don't think this matters.

tripodsan:
> Of course, you can set the property explicitly in the function definition

I would avoid this, too. so do you revert the changes like `fetch.ext = 'fetch';` ?

kptdobe:
If this is the answer to https://github.com/adobe/developer.adobe.com-planning/issues/148, could you write a test that illustrate how this would be done (i.e. show an example how to support a markdown extension) ?
We need to make sure this would be doable and this is not too complicated.

Also, how do I "replace" the existing function ?  ;)

As @tripodsan stated, those `before` / `after` extension points introduces an uncontrollable API which will at one point break upgrades and backward compatibility. And here, we have a bigger issue which is revealed by this PR: each function does not have a precise and strict contract! At the moment, we only need one function, the `pre`, which is more than enough.

If you look at each function (like the `getmetada` https://github.com/adobe/helix-pipeline/blob/461d3f64f48d3c055b304dda627c4199714ef207/src/html/get-metadata.js#L145), you could do whatever you want with the payload, only contract is that it must return the payload. Why would I do a `before` / `after` `meta` since I can do the same in the `after` `fetch` ? 
The `getmetada` does "something" with the sections which are required later but this is not clear in the contract and never validated.

trieloff:
@kptdobe I'm working on a realistic example for 
https://github.com/adobe/developer.adobe.com-planning/issues/148 in https://github.com/adobe/helix-cli/pull/653 (although I'm currently blocked by #159)

> Also, how do I "replace" the existing function ? ;)

You don't, but you can make sure that the preconditions for skipping a function are met (see the 404 example).

> Why would I do a before/after `meta` since I can do the same in the after `fetch`?

Because you either want to satisfy a pre-condition of `meta` or work with data generated by `meta`. 

You may be able to do the precondition handling right after `fetch`, but I think it's a good practice to defer.

If strictness of the contract is a concern for you, we can introduce function-specific pre- and post-schemas.

@tripodsan 

> revert the changes like `fetch.ext = 'fetch';`

- [ ] Will do.

src/pipeline.js
trieloff: Done

PR #236, opened 2019-04-04
openwhisk.js to add request body to payload

### LINKED ISSUE #235: openwhisk emulator should pass request body to pipeline
If there is a `body.content` in the `actionParams`, add it to the `payload` (like openwhisk would). Depends on https://github.com/adobe/helix-simulator/issues/179.

rofe:
Fixes #235

### DIFF
--- a/src/utils/openwhisk.js
+++ b/src/utils/openwhisk.js
@@ -133,7 +133,7 @@ async function runPipeline(cont, pipe, actionParams) {
       request: extractClientRequest(action),
     };
     // if request has JSON body.content, pass it to payload.content
-    if (actionParams.body && action.request.headers['content-type'].endsWith('json')) {
+    if (actionParams.body && action.request.headers['content-type'] === 'application/json') {
       try {
         const body = JSON.parse(actionParams.body);
         if (body.content) {
--- a/src/utils/openwhisk.js
+++ b/src/utils/openwhisk.js
@@ -90,6 +90,7 @@ function extractActionContext(params) {
     __ow_headers,
     __ow_method,
     __ow_logger,
+    body, // skip body here
     ...disclosed
   } = params;
 
@@ -131,6 +132,17 @@ async function runPipeline(cont, pipe, actionParams) {
     const payload = {
       request: extractClientRequest(action),
     };
+    // if request has JSON body.content, pass it to payload.content
+    if (actionParams.body && action.request.headers['content-type'].endsWith('json')) {
+      try {
+        const body = JSON.parse(actionParams.body);
+        if (body.content) {
+          payload.content = body.content;
+        }
+      } catch (e) {
+        action.logger.warn('Invalid request body found', e);
+      }
+    }
     return pipe(cont, payload, action);
   }
   return createActionResponse(await owwrapper(runner, actionParams));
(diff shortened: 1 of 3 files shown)


tripodsan:
Quick review: 
- check content type correctly.
- use `request.body` instead

rofe:
> * check content type correctly.

Ack. I wanted to be smart and allow `*json`, but i guess `application/json` is the valid content type to go for.

> * use `request.body` instead

which `request.body` do you mean? The body of the original request doesn't seem to get passed anywhere else, and i also don't see a `body` in the `request.schema.json` or `rawrequest.schema.json`...

test/testOpenwhisk.js
```diff
@@ -140,4 +140,44 @@ describe('Testing OpenWhisk adapter', () => {
...
+      __ow_headers: {
+        'content-type': 'application/json',
+      },
+      __ow_logger: log,
+      __ow_method: 'post',
+    };
+    let payload = {};
+
+    // adds JSON from request body.content to payload.content
+    params.body = '{ "content": { "body": "# Foo" } }';
```
tripodsan: openwhisk already invokes the action with the json decoded.

src/utils/openwhisk.js
```diff
@@ -90,6 +90,7 @@ function extractActionContext(params) {
     __ow_headers,
     __ow_method,
     __ow_logger,
+    body, // skip body here
```
tripodsan: I would keep it there. in case the `body` doesn't provide the `content.body` but other data that might be useful.

src/utils/openwhisk.js
```diff
@@ -131,6 +132,17 @@ async function runPipeline(cont, pipe, actionParams) {
     const payload = {
       request: extractClientRequest(action),
     };
+    // if request has JSON body.content, pass it to payload.content
+    if (actionParams.body && action.request.headers['content-type'] === 'application/json') {
+      try {
+        const body = JSON.parse(actionParams.body);
```
tripodsan: you don't need to parse it, as openwhisk already provides the `json` body as action param.

src/utils/openwhisk.js
```diff
@@ -131,6 +132,17 @@ async function runPipeline(cont, pipe, actionParams) {
...
+        if (body.content) {
+          payload.content = body.content;
```
tripodsan: I would set the `body` directly to the content, eg:

`payload.content = body`

test/testOpenwhisk.js
tripodsan: also, this should probably just be `params.body = '# Foo';`

as openwhisk promotes the json properties as params. for example:

```
$ curl https://${APIHOST}/api/v1/web/guest/demo/hello.json -H 'Content-Type: application/json' -d '{"name":"Jane"}'
{
  "response": {
    "name": "Jane",
    "__ow_method": "post",
    "__ow_headers": {
      "accept": "*/*",
      "connection": "close",
      "content-length": "15",
      "content-type": "application/json",
      "host": "172.17.0.1",
      "user-agent": "curl/7.43.0"
    },
    "__ow_path": ""
  }
}
```

src/utils/openwhisk.js
```diff
@@ -131,6 +132,17 @@ async function runPipeline(cont, pipe, actionParams) {
     const payload = {
```
tripodsan: can we rename `payload` to `context` to avoid further misnaming ?

rofe:
Will no longer be implemented, see conclusion in #235.

PR #243, opened 2019-04-09
Process content from post request

### LINKED ISSUE #235: openwhisk emulator should pass request body to pipeline

rofe:
Fixes #235


src/utils/openwhisk.js
```diff
@@ -90,6 +90,7 @@ function extractActionContext(params) {
     __ow_headers,
     __ow_method,
     __ow_logger,
+    content,
```
tripodsan: I'm not sure if we should remove it. we don't remove the other params neither....

In any case: does't lint complain that there is an unused variable? maybe better to delete it:

```
delete disclosed.content;
```

tripodsan:
please  remove `package-lock.json` from your PR...

src/utils/openwhisk.js
rofe: I think `content` is a special case compared to other params. lint didn't complain, but I'll `delete` it anyway.

PR #263, opened 2019-04-15
fix(html pipe): Sanitize generated markdown to avoid XSS attacks

### LINKED ISSUE #253: [XSS] Sanitize generated markdown to avoid XSS attacks
The Links Markdown is easily escapable

either 

`[I'm an inline-style link](javascript://%0Dalert(document.domain))`

or

`[I'm an inline-style link](javascript:alert(document.domain))`

produces a valid XSS payload (after clicking on the produced link)

ramboz:
Properly escape the generated HTML using DOMPurify to avoid potential XSS attacks via JS injection, and also avoid DOM clobbering.

- Add an `.before` step right before the `pre` function in the html pipeline to sanitize the content
- Adjust the default DOMPurify sanitization config to support the helix use case (`esi:` tags)
- Add unit tests to validate escaping works on links and images
- Add unit tests to validate custom elements and attributes in the MD are kept
- Add unit tests to validate sanitization only applies to author-generated content and not the developer content (i.e. HTL template)
- Add ADR to document the decision

fix #253


src/utils/sanitize-hast.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+const _ = require('lodash');
+const sanitize = require('hast-util-sanitize');
+const GITHUB_SCHEMA = require('hast-util-sanitize/lib/github');
+
+// Define helix specific customizations on top of the default GitHub schema
+const HELIX_SCHEMA_CUSTOMIZATIONS = {
+  tagNames: ['esi:include'],
+  attributes: {
+    'esi:include': ['src'],
```
tripodsan: `esi:include` ? this is never part of the markdown - or at least, it shouldn't be.
so best to remove it completely.

test/testMdastToVDOM.js
```diff
@@ -65,10 +65,10 @@ describe('Test MDAST to VDOM Transformation', () => {
   it('Programmatic Matcher Function', () => {
     const mdast = fs.readJSONSync(path.resolve(__dirname, 'fixtures', 'simple.json'));
     const transformer = new VDOM(mdast, action.secrets);
-    transformer.match(({ type }) => type === 'heading', () => '<h1>All Headings are the same to me</h1>');
+    transformer.match(({ type }) => type === 'heading', () => '<h1 id="h1">All Headings are the same to me</h1>');
```
tripodsan: is this relevant to the XSS changes?

test/testMdastToVDOM.js
```diff
@@ -65,10 +65,10 @@ describe('Test MDAST to VDOM Transformation', () => {
...
     assertTransformerYieldsDocument(
       transformer,
-      '<h1>All Headings are the same to me</h1>',
+      '<h1 id="user-content-h1">All Headings are the same to me</h1>',
```
tripodsan: is this relevant to the XSS changes?

test/testMdastToVDOM.js
```diff
@@ -79,7 +79,7 @@ describe('Test MDAST to VDOM Transformation', () => {
     assertTransformerYieldsDocument(
       transformer, `
       <div>
-        <a name="h1"></a>
+        <a name="user-content-h1"></a>
```
tripodsan: is this relevant to the XSS changes?

test/testMdastToVDOM.js
trieloff: Seems like a leak from your other branch.

src/utils/mdast-to-vdom.js
```diff
@@ -194,7 +195,9 @@ class VDOMTransformer {
    * @returns {string} the corresponding HTML
    */
   static toHTML(mdast, handlers) {
-    return hast2html(mdast2hast(mdast, { handlers }));
+    const rawHast = mdast2hast(mdast, { handlers });
+    const sanitizedHast = sanitize(rawHast);
```
koraa: There are a ton of well maintained, well reviewed (just by sheer amount) html sanitizers out there. Just fom 5s of research I get:

https://github.com/cure53/DOMPurify

I would rather trust one of those – if just by sheer number of users – than a 
library that is used by few and touched every couple of months…
This is a security issue after all…

I don't see an advantage of doing this in htast over dom – but I am tooting the same horn here as in the other PR. If we model this as a dom transforming pipeline step we give users the ability to opt out (maybe they want to have js in ther markdown) and we are Independent from htast/remark.

test/testMdastToVDOM.js
ramboz: It actually is related to XSS. This is to avoid DOM clobbering.
Both `id` and `name` are exposed on the global window object, and if you have a non-secure browser (webkit is, but some other aren't), then you can actually override global properties on the window with this.

```
<h1 id="addEventListener"></h1>
<a name="removeEventListener"></h1>
```

The best practice, and also what is used by GitHub, is to always prefix these 2 properties with "user-content-" when they are coming from user input (i.e. the markdown in our case).

test/testMdastToVDOM.js
ramboz: Same as above. This is the DOM clobbering escape

test/testMdastToVDOM.js
ramboz: It is not necessarily needed, but was an easy place to test the DOM clobbering escape. I could duplicate the test otherwise to cover that use case

src/utils/sanitize-hast.js
```diff
@@ -0,0 +1,46 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+const _ = require('lodash');
+const sanitize = require('hast-util-sanitize');
+const GITHUB_SCHEMA = require('hast-util-sanitize/lib/github');
+
+// Define helix specific customizations on top of the default GitHub schema
+const HELIX_SCHEMA_CUSTOMIZATIONS = {
+  tagNames: ['esi:include'],
```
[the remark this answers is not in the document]
ramboz: Agreed. But then again, white-listing is the safest XSS approach.
The best security is always "deny-all" and open up what you need. I'd rather let the customer extend the schema with their own as needed, than just "allow-all".

What do you think?

src/utils/mdast-to-vdom.js
ramboz: @koraa 

> than a library that is used by few and touched every couple of months…

I'd tend to agree here. Although this was updated in the last couple months and does have some good tests for it. 

Plus it is from the same org as a lot of our other modules (hast-to-hyperscript, hast-util-to-html, mdast-util-to-hast, mdast-util-to-string, unist-util-map, unist-util-select, …) which basically face the same issues, so my assumption was these thresholds (last commit, number of maintainers, tests, etc.) were the accepted norm.

I'd be happy to start a discussion about how to properly chose dependencies for helix. Since the project is still in its infancy, it's a good time to define best practices and choose dependencies carefully for the long run.
Would you agree to handle that in a separate issue and close the security hole for now with this?

koraa: Yeah, but those other libraries are not as security-relevant I would say…

It does have good tests, that's true. I personally do not have specific thresholds for anything; it's more about comparing the options and finding the best choice.
How about you take a look at a couple of html sanatization libs and compare them?

koraa: I also tend to take a look at the source code of libs and tbh I am more impressed with  DOMPurify than hast-util-sanitize (the latter uses a bit of a weird style and the former seems to be much better documented and more complete)

Btw I didn't mean to imply that hast-util-sanitize was a bad or amateurish choice; if it was the only choice we had, I'd go for it. But it isn't the only choice and I think we may be able to do better.

ramboz: Agree. That is also why I hope to get some more input from @asanso. I don't want to make a pure judgment call on the repo itself for now, and want to make sure we properly cover the XSS use case with the library.

I'm also not partial with `hast-util-sanitize`. I like the kind of discussion this PR triggers.
I'll play around with other libs this week and see what I can come up with

ramboz:
@koraa I see your point. I agree at the moment that the risk coming from the author is quite limited.

Now if I try to imagine a worst case scenario, I'd be looking at the risk of having someone gaining access to the author account, or being added to the project team, and then maliciously injecting random JS in the markdown (bitcoin miner, keylogger, etc.). Also, just a non-savvy author could randomly copy-paste from stackoverflow and end up opening some loopholes.

I could also imagine that at some point we would also integrate git comments, etc. in the pipeline and this would open up additional attack vectors. So I think sanitization and XSS escaping is definitely needed.

koraa:
> I could also imagine that at some point we would also integrate git comments, etc. in the pipeline and this would open up additional attack vectors. So I think sanitization and XSS escaping is definitely needed.

Agreed!

src/utils/mdast-to-vdom.js
koraa: ++1

tripodsan:
> @lkrapf mentioned that the HTL regexp in Sling evolved quite a bit since you ported it to node.

why regexp?

lkrapf:
@tripodsan wrote:
> why regexp?

This is the regexp Sling uses to validate href's:
https://github.com/apache/sling-org-apache-sling-xss/blob/5071a142cdfeaf128a0db1ed5222b4e3753303c9/src/main/resources/SLING-INF/content/config.xml#L76

built from this code:
https://github.com/apache/sling-org-apache-sling-xss/blob/5071a142cdfeaf128a0db1ed5222b4e3753303c9/src/main/java/org/apache/sling/xss/impl/XSSFilterImpl.java#L84

However, a good alternative for Helix might be to use https://github.com/cure53/DOMPurify, it is very well test it, and you could use it for both client- and serverside.

src/utils/heading-handler.js
```diff
@@ -55,7 +55,7 @@ class HeadingHandler {
       // Inject the id after transformation
       const n = Object.assign({}, node);
       const el = fallback(h, n);
-      el.properties.id = headingIdentifier;
+      el.properties.id = el.properties.id || `user-content-${headingIdentifier}`;
```
ramboz: If an `id` was provided, keep it as is, otherwise make sure to escape it to avoid DOM clobbering.
If provided `id` does clobber the DOM, DOMPurify will remove it during the sanitization step.

test/testEmbedHandler.js
```diff
@@ -115,6 +115,6 @@ https://www.youtube.com/watch?v=KOxbO0EI4MA
     assert.equal(result.response.body, `<p>Hello World
 Here comes an embed.</p>
 <esi:include src="https://example-embed-service.com/https://www.youtube.com/watch?v=KOxbO0EI4MA"></esi:include>
-<p><img src="easy.png" alt="Easy!" srcset="easy.png?width=480&amp;auto=webp 480w,easy.png?width=1384&amp;auto=webp 1384w,easy.png?width=2288&amp;auto=webp 2288w,easy.png?width=3192&amp;auto=webp 3192w,easy.png?width=4096&amp;auto=webp 4096w" sizes="100vw"></p>`);
```
ramboz: No particular changes here. DOMPurify just reorders the properties in the output

test/testMdastToVDOM.js
ramboz: Removed as I now have dedicated tests for XSS and DOM Clobbering

ramboz:
@tripodsan, @rofe: not sure about the HTL escape. It didn't seem to work here.
For now I updated the PR to use DOMPurify instead and also sanitize the output instead of the MDAST

rofe:
`htlengine` is supposed to output safe HTML in the first place. On the other hand, it's probably a good idea to run the response.body through a proven sanitizer before sending it to the browser.

tripodsan:
> `htlengine` is supposed to output safe HTML in the first place.

maybe, the problem is different. since HTL is designed as a markup language where the HTL itself is the primary source of the HTML. the content is usually just injected into the attributes or the node text.

in the pipeline he have another setup, where the majority of the markup is generated via the DOM when injecting it into the output via `${content.document.innerHTML)` or similar. I'm not sure if we can filter out all XSS efficiently in HTL, w/o compromising intentional javascript links. but maybe I'm wrong.

> On the other hand, it's probably a good idea to run the response.body through a proven sanitizer before sending it to the browser.

maybe. but wouldn't that invalidate clickable links, like `<a href="#" onclick="doSomething" >` ?

src/html/sanitize.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+const createDOMPurify = require('dompurify');
+const { JSDOM } = require('jsdom');
+
+const helixSanitizationConfig = { ADD_TAGS: ['esi:include'] };
+
+function sanitize({ response }, { logger }) {
+  logger.log('debug', 'Sanitizing reponse body to avoid XSS injections.');
+
+  const { window } = (new JSDOM(''));
```
tripodsan: I wouldn't call this `window`. maybe `globalContext` or something?

test/testEmbedHandler.js
koraa: There is a domEquivalence testes in helix.shared/src/dom.js that is precisely designed to work in such situations.

koraa:
@rofe @tripodsan Just wondering, how can htlengine output `safe` html? AFAICS there is no catch-all solution for preventing XSS attacks since such a solution would have to remove all javascript tags and we do want developers to be able to embed js in their sites?

Maybe we should take a step back and decide on which specific attacks we are trying to guard against and which we are not trying to guard against?

koraa:
Does htl have xss prevention helpers btw?

koraa:
https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md
https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.md

ramboz:
@koraa HTL does XSS escaping automatically for us based on the [display context](https://docs.adobe.com/content/help/en/experience-manager-htl/using/htl/expression-language.html#display-context)
But it's not necessarily a catch-all solution as it was designed with AEM components development in mind. Also, it's still not clear to me whether HTL will always be used as a template engine in helix, or if that engine is supposed to be plug-and-play… in which case we need a sanitization step at the end anyway, right?

koraa:
https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md
https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.md

I can definately inject `javascript:` links into the markdown. @helix-pipeline crashes (see below) but if I fix that bug, I can inject them.

```
TypeError: Expected input to be a string
    at Function.picomatch.test (/home/karo/pr0j/adobe/helix-cli/node_modules/picomatch/lib/picomatch.js:111:11)
    at matcher (/home/karo/pr0j/adobe/helix-cli/node_modules/picomatch/lib/picomatch.js:59:48)
    at items.some.item (/home/karo/pr0j/adobe/helix-cli/node_modules/@adobe/helix-pipeline/node_modules/micromatch/index.js:258:28)
    at Array.some (<anonymous>)
    at Function.micromatch.some (/home/karo/pr0j/adobe/helix-cli/node_modules/@adobe/helix-pipeline/node_modules/micromatch/index.js:258:15)
    at embed (/home/karo/pr0j/adobe/helix-cli/node_modules/@adobe/helix-pipeline/src/html/find-embeds.js:55:10)
    at map (/home/karo/pr0j/adobe/helix-cli/node_modules/@adobe/helix-pipeline/src/html/find-embeds.js:73:7)
    at preorder (/home/karo/pr0j/adobe/helix-cli/node_modules/unist-util-map/lib/unist-util-map.js:13:30)
    at bound (/home/karo/pr0j/adobe/helix-cli/node_modules/unist-util-map/lib/unist-util-map.js:22:14)
    at Array.map (<anonymous>)     
```

src/defaults/html.pipe.js
```diff
@@ -58,6 +59,7 @@ const htmlpipe = (cont, payload, action) => {
     .before(html).expose('html')
     .before(responsive)
     .once(cont)
+    .after(sanitize)
```
tripodsan: this is a wrong place. at this moment, the HTML is already written to the response.body. the `cont` is the step that executes the template engine.

tripodsan:
> I can definately inject javascript: links into the markdown. @helix-pipeline crashes (see below) but if I fix that bug, I can inject them and they are not removed by htl…

@koraa yes, this is the original bug #253 :-)

As I said, HTL is meant as template language for HTML. so all code that is directly in the `.htl` file, doesn't need XSS escaping, because it's _you_ that types it. for all content that is sourced from variables, the XSS escaping based on the [context](https://github.com/adobe/htl-spec/blob/master/SPECIFICATION.md#121-display-context). for example if you have a `<a href="${mylink}"/>` then it assumes that the `mylink` is user generated, and applies the escaping suitable for attribute values. of you have `<div>${mytext}</div>` then it assumes user generated HTML and escapes accordingly. if you also want to escape all HTML, you could write `<div>${mytext @ context="text"}</div>`.

In our case, we have a mixed scenario, because we want to inject the HTML from the JDOM directly, but also be safe against XSS. so we can't just apply the normal escaping of the HTL engine, since it would remove too much from the output. so I believe using DOMPurify on the dom is the best way to protect against XSS in the markdown, and still have the flexibility for pipeline steps to add more sophisticated nodes.

> Also, it's still not clear to me whether HTL will always be used as a template engine in helix, or if that engine is supposed to be plug-and-play…in which case we need a sanitization step at the end anyway, right?

@ramboz no, HTL is just 1 possible way to generate markup. but I think that XSS prevention should mainly happen against any (user generated) content. so applying it to the DOM before the dom is render by the template engine is IMO correct.

src/defaults/html.pipe.js
tripodsan: forget about the comment above. I thought that the purify would operate on the VDOM and not on the html itself. so in fact, it **is** reparsing the entire output...

tripodsan: so I'm no longer sure that this is correct. I would rather have just sanitized the VDOM before it is used by HTL. otherwise you remove too much power from the templating....

ramboz: 👍 I see your point. DOMPurify should be able to work on the VDOM, so I'll refactor accordingly if there are no objections from the others

rofe: But if the sanitization takes place before the `cont`, wouldn't we run the risk of developers introducing potential security issues? I think that parsing the final output would be the safer bet...

tripodsan: > wouldn't we run the risk of developers introducing potential security issues? 

that's why they use HTL to minimize this :-) but I think they should still have the power to write things like `<a href="#" onclick="..."/>` , if they want.

I think we should have more examples, of what should be possible and what to guard against.

trieloff:
I think there are three levels of trust:

1. *burn it with fire*: user-generated content
2. *people make mistakes*: author-generated content
3. *she probably knows what she's doing*: developer-generated markup

In this PR (which maybe should go back to being an issue) we are tacking (2) and it should be done on the DOM that is being passed into the pre function. Maybe even on the MDAST, although the available filtering tools won't be as solid.

ramboz:
Alright, so to summarize the whole discussion, in this PR the scope is to:
- sanitize the author-generated content (i.e. the markdown)
- not sanitize the developer-generated markup (i.e. the HTL template)
- we want to perform that operation as part of the HTML Pipeline, right before the `pre` function is executed

Which means that for the following example:

`index.md`
```markdown
# Foo
[Bar](javascript:alert('XSS1'))
```

`html.htl`
```html
<html>
  <head>
  </head>
  <body>
    ${content.document.body.innerHTML}
    <a href="javascript:alert('XSS2')">Baz</a>
  </body>
</html>
```

We'll expect this output:
```html
<html>
  <head>
  </head>
  <body>
    <h1>Foo</h1>
    <a>Bar</a>
    <a href="javascript:alert('XSS2')">Baz</a>
  </body>
</html>
```

src/utils/heading-handler.js
trieloff: The `user-content-` prefix you are inserting here seems odd and is leading to heading IDs that don't match GitHub's.

ramboz: Github does the same. You can check the helix home readme for instance and you'll notice that
```markdown
# Background
```
becomes
```html
<h2>
    <a id="user-content-background" class="anchor" aria-hidden="true" href="#background"></a>
    Background
</h2>
```

ramboz:
As explained in the comments, this is to avoid DOM clobbering (which is another for of attack that can be combined with XSS).

src/utils/heading-handler.js
trieloff: I didn't know that the actual IDs in GitHub are different and that GitHub is using JavaScript to restore the scroll functionality they've broken through this clobbering.

I think we should not handle it the same way. Use proper `id` attribute values, and maybe add a blacklist of known DOM properties that are forbidden. You have access to JSDOM, which could help.

ramboz: It's a tricky one to handle, as white/black listing is not necessarily enough, you also have to account for all character encoding possibilities. Any value put in an `id` or `name` attribute can potentially clobber the DOM, and browsers are _supposed_ to prevent that (Chrome does, Firefox has a long history of not doing it, …)

If you only blacklist, then you fail to escape new properties that are introduced by the browser engine. The accepted norm here is to prefix using `user-content-` like github does it.

ramboz: By the way, if I let these through then DOMPurify just completely removes the identifiers, but then you can't target them in the URL either. In most cases, this is maybe not a problem, but the moment you start documenting APIs on a public website, chances are high that you use property names for the headings, and then you mostly end up with no heading identifiers at all.

rofe:
> `html.htl`
> ```
>     <a href="javascript:alert('XSS2')">Baz</a>
> ```
off topic: i'd argue that we should never let HTL output something like that, even if the developer insists. but that's an `htlengine` topic.

ramboz:
@rofe I'm with you on that one. See https://github.com/adobe/helix-pipeline/pull/263#issuecomment-485449022. I would rather never have any JS in the DOM if possible

src/utils/heading-handler.js
trieloff: The potential attack vector is relatively small, as it would require
- a new DOM property
- that is not yet reflected in JSDOM
- that does not use camel-case syntax (our slugging makes everything lowercase)

Maybe enabling this prefix for single-word headlines only would be sufficient, as JS does not allow properties with a dash in them.

trieloff: Seems to me that the cure is worse than the disease.

ramboz: > Seems to me that the cure is worse than the disease.

Seems to me the this is a subjective argument and we could argue for hours on that ;)
Security experts seem to think otherwise if prefixing has been accepted as the best practice.

I'd rather have a security expert take the decision on that one. I'm not familiar enough with the topic to be able to judge what all the possible edge cases are for DOM Clobbering. Various articles I found seem to show it's quite easy to write them, so I'd be leaning towards the safest approach, that a custom one that we do not fully understand.

Also, in my previous PR on the headings, the ask was to behave exactly like github does, and here your ask is to not do so… I'm a bit confused.

trieloff: > Also, in my previous PR on the headings, the ask was to behave exactly like github does, and here your ask is to not do so… I'm a bit confused.

I was mistaken. I assumed that GitHub used the IDs that appeared when you hovered over the link symbol in a headline and took that for face value instead of inspecting the HTML.

GitHub's approach would require loading JavaScript into every customer site, just to get links to fragment identifiers to work as expected. This is not acceptable.

It looks like at least three static site generators I've looked at decided that this is not a problem:
- https://gohugo.io/templates/lookup-order/#hugo-layouts-lookup-rules
- https://jekyllrb.com/docs/#instructions
- https://www.gatsbyjs.org/tutorial/#advanced-tutorials

(note the fragment identifiers in the URL, which all work, and are all in the HTML, and are all not dangerous.)

ramboz: > GitHub's approach would require loading JavaScript into every customer site

Not necessarily. You can also achieve this by overriding the heading output and injecting the anchors and accepting that your has includes the prefix.
At the moment we only output the identifier, not the anchor to target it, so any site will have to add some logic on top. Whether it's clientisde JS or using a matcher function to extend the heading markup is probably a project decision.

> It looks like at least three static site generators I've looked at decided that this is not a problem:

I'd have to write some example on their stack to see how they really handle the issue:
- do they ignore it altogether?
- do they just remove the offending id/name like DOMPurify does?
- do they have another method?

The decision is also ultimately dependent on who the producer of the markup is?
DOM clobbering mostly requires a conscious decision to "hack" the content, so if we only look at authors writing content, the risk is mostly limited. But the moment you have a use case where the community is creating content via comments for the site, then you need more security.

I'd be fine ignoring the problem for now (as DOMPurify anyway cleans the id/name), and maybe documenting the limitation somewhere, if we only consider controlled authors to be the content generators.

trieloff: The markup producer is the author and trusted. I'd assume that most DOM clobbering issues that arise would be accidents rather than attacks.

src/utils/sanitize-hast.js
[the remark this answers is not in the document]
trieloff: I don't know if this is still an issue with your new implementation. For normal HTML elements, let's use a whitelist of known good elements. For custom elements, let's assume that they are all good, because they would have to be inserted by the developer of the site.

As long as we don't have a good way of configuring the XSS filter (and we probably should not introduce a configuration anyway), we should have defaults that work, even when you are using extensions.

src/utils/heading-handler.js
ramboz: Ok, so assuming an "accident":
```
## Name
```

The process will look like this:
- generate a heading with identifier `name`
- DOM gets sanitized by DOMPurify which detects a collision on `document.name`, and removes it
- final output is:
```
<h2>Name</h2>
```
with no anchor.

If customers want the prefixed version, they can do that in their own matching function for the heading

src/utils/sanitize-hast.js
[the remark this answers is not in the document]
ramboz: It still would be. DOMPurify will remove any custom elements by default.
We'd have to follow their proposed solution to https://github.com/cure53/DOMPurify/issues/184 in:
https://github.com/cure53/DOMPurify/blob/master/test/test-suite.js#L294-L312

> we probably should not introduce a configuration anyway

I'm not sure about that one. We can have a catch-all solution for `<\w+-\w+` tags, but you'd still end up with edge cases that customers may need to have.
- What about `<customtag/>`?
- Should we support custom properties as well? `<custom-tag custom-prop/>`
- What about `<esi:custom/>`?

Assuming all custom elements are "good" is the same as saying all authors know what they are doing and their sessions cannot be compromised. While I'm ok with that argument for DOM Clobbering topic as DOMPurifies is applied after that anyway and secures the response, I'm really not convinved here.

My proposal would be to "white-list" all the `esi:` tags we need, but lock down the rest and let it be configured for each project by extending the default DOMPurify config (and abstracting it obviously as it's an implementation detail).

Just a random thought, but I would be happy as a front-end dev to define something like:
`.helix-config.json`
```js
{
  "allow-custom-html-tags": true, // allow all
  "allow-custom-html-tags": ["my-tag", "my-other-tag"], // allow some
  "allow-custom-html-attributes": true, // allow all
  "allow-custom-html-attributes": ["my-attribute", "my-other-attribute"], // allow some
}
```
i.e. the same I would define my linting configs for instance.

src/utils/heading-handler.js
trieloff: Looks good. This is assuming that

```markdown
## Name 1
```

would get an `id`, and `DOMPurify` isn't simply removing all `id` attributes.

src/utils/sanitize-hast.js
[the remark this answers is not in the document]
trieloff: > What about `<customtag/>`?
not a Custom Element, according to [spec](http://w3c.github.io/webcomponents/spec/custom/#prod-potentialcustomelementname): forbidden.
> Should we support custom properties as well? `<custom-tag custom-prop/>`
within a custom element, yes.
> What about `<esi:custom/>`?

Neither a proper Custom Element, nor part of the ESI language (https://www.w3.org/TR/esi-lang): forbidden

For ESI, we need a while list:
* `esi:include`
* `esi:remove`

For [Custom Elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) we don't know what's coming in advance, so we can assume that the developer who is building the VDOM Matcher to produce `my-foo-element` is not trying to attack the other developer, working on the same code base, who is implementing `class MyFooElement extends HTMLParagraphElement…` on the client-side.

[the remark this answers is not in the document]
ramboz: I generally agree. But I'd really like to see some potential attack vectors described and tests for them being written. At the moment, I can't come up with any, but maybe @lkrapf or @asanso can?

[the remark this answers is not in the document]
trieloff: I'd only allow the ESI tags that are actually supported by Fastly (complete list above)

src/utils/heading-handler.js
ramboz: Yes,
```markdown
## Name 1
```
would be kept and turned into:
```html
<h2 id="name-1">Name 1</h2>
```

In the end, I only see this being a real issue on API documentation websites where the author would probably use the API name for the heading:
```markdown
# My Custom Object
## Properties
### Name
### OnClick
### OnChange
…
```

And for these edge cases, I would either:
- Add a prefix to all headings: `Property: Name`, `Event: OnClick` in the markdown
- Use a custom matcher function that injects a prefix in the identifier in the HTML
- Use a custom JS script that does it clientside

@trieloff, @kptdobe, @rofe:
Since this would be a known edge case, where should we document this?
Is helix using ADRs (https://adr.github.io/) or do we have customer-facing documentation that would need an entry for this?

src/utils/heading-handler.js
```diff
@@ -45,7 +45,7 @@ class HeadingHandler {
       // Inject the id after transformation
       const n = Object.assign({}, node);
       const el = fallback(h, n);
-      el.properties.id = headingIdentifier;
+      el.properties.id = el.properties.id || headingIdentifier;
```
ramboz: Just keep existing identifiers if defined by the author (not really possible in markdown for now, but if we use asciidoc as a parser at some point, then it would be supported

src/utils/heading-handler.js
trieloff: We don't use ADRs, but it looks like a good template for this, so just put the first ADR into docs. Removing the ID for `properties`, `name`, etc. only seems like a good solution and there are enough documented workarounds that customers can implement if needed.

ramboz:
@trieloff You still have a request for change open on this PR. What's the process in this case for the team? Just I just go ahead and ignore since the changes were tackled, or do you consider this veto-ing the PR until it is properly re-reviewed and approved?

trieloff:
@ramboz please resolve the conflict and re-request the review. As far as I can see, the conflict has been the main thing keeping me from merging it.

ramboz:
@trieloff I've been rebasing a couple times to resolve the conflict. I'll do it again!

ramboz:
@trieloff Fixed the merge conflict and updated the tests according to the latest changes. Please take a moment to re-review, thanks :)

docs/adr/heading-identifiers.md
```diff
@@ -0,0 +1,39 @@
...
+In the context of generating heading identifiers for easier URL targeting,
+facing the concern that some of them could end up clobbering the DOM
+we decided to sanitize the offending identifiers
+to achieve proper protection of the DOM,
+accepting that in limited use cases where the autor would need to expose a heading that clobbers, that heading would just not get an identifier.
+
+Context
+-------
+
+> Describe the detailed context and problem being addressed
```
tripodsan: line break missing

```suggestion
> Describe the detailed context and problem being addressed

```

src/defaults/html.pipe.js
```diff
@@ -63,6 +64,7 @@ const htmlpipe = (cont, payload, action) => {
     .before(selecttest)
     .before(html).expose('html')
     .before(responsive)
+    .before(sanitize)
```
tripodsan: once we modified the entire pipeline (including the HTL engine) to be based on JSDOM (see #337), should be move this after the `once()`  step?

trieloff: No, because this would make it impossible to have unsanitized output, which would severely reduce the power and expressiveness of the templates. As we generally assume developers know what they want, we shouldn't do this.

trieloff:
Thanks for your contribution and your patience with us @ramboz

PR #289, opened 2019-04-25
feat(json): refactoring json pipeline

### LINKED ISSUE #230: JSON pipeline should emit JSON based on a (dedicated) `content` property by default
Following up on a previous [discussion](https://github.com/adobe/helix-pipeline/issues/218#issuecomment-475789765), I think the JSON pipeline developer flow diverges from the HTML and XML pipeline:
* The HTML pipe has an `emit` step which writes HTML to the `response.body` from `content.document` (which is generated out of the box, can be manipulated in the `pre` function).
* The XML pipe has an `emit` step which writes XML to the `response.body` from `content.xml` (which the developer needs to populate in the `main` function, because we can't guess what the default output should be)
* The JSON pipe does not use a dedicated property in `content`, nor does it write any output by default. Instead it relies on the `main` function to do so.

I think this is inconsistent. Let's refactor the JSON pipe to have an `emit` (and subsequent `check`, `cache`*, `status`) step which transforms `content.json` (provided by the developer in `main`) into JSON and writes it into `response.body`.

cc @trieloff, @kptdobe

### LINKED ISSUE #269: Add extension points to JSON pipeline
Similar to the HTML pipeline, the JSON pipeline should expose its own extension points.

### LINKED ISSUE #280: JSON pipeline should not display error in production
Like the HTML and XML pipeline, the JSON pipeline should only output an error message in a non-production environment.

rofe:
- fix #230: emit json from content.json
- fix #269: expose extension points
- fix #280: don't bleed errors in production

### DIFF
--- a/src/defaults/json.pipe.js
+++ b/src/defaults/json.pipe.js
@@ -16,33 +16,38 @@ const fetch = require('../html/fetch-markdown.js');
 const parse = require('../html/parse-markdown.js');
 const meta = require('../html/get-metadata.js');
 const type = require('../utils/set-content-type.js');
-const smartypants = require('../html/smartypants');
-const sections = require('../html/split-sections');
-const production = require('../utils/is-production');
+const smartypants = require('../html/smartypants.js');
+const sections = require('../html/split-sections.js');
+const production = require('../utils/is-production.js');
 const dump = require('../utils/dump-context.js');
-const validate = require('../utils/validate');
-const parseFrontmatter = require('../html/parse-frontmatter');
+const validate = require('../utils/validate.js');
+const emit = require('../json/emit-json.js');
+const { selectStatus } = require('../json/set-json-status.js');
+const parseFrontmatter = require('../html/parse-frontmatter.js');
 
 /* eslint no-param-reassign: off */
+/* eslint newline-per-chained-call: off */
 
-const htmlpipe = (cont, payload, action) => {
+const jsonpipe = (cont, payload, action) => {
   action.logger = action.logger || log;
   action.logger.log('debug', 'Constructing JSON Pipeline');
   const pipe = new Pipeline(action);
   pipe
     .every(dump).when(() => !production())
     .every(validate).when(() => !production())
-    .before(fetch)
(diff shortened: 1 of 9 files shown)


src/defaults/json.pipe.js
```diff
@@ -16,33 +16,38 @@ const fetch = require('../html/fetch-markdown.js');
...
+const validate = require('../utils/validate.js');
+const emit = require('../json/emit-json.js');
+const { selectStatus } = require('../json/set-json-status.js');
+const parseFrontmatter = require('../html/parse-frontmatter.js');
 
 /* eslint no-param-reassign: off */
+/* eslint newline-per-chained-call: off */
 
-const htmlpipe = (cont, payload, action) => {
+const jsonpipe = (cont, payload, action) => {
```
trieloff: Good catch.

src/json/emit-json.js
```diff
@@ -0,0 +1,35 @@
...
+  if (response.body) {
+    logger.debug('Response body already exists');
+    return {};
+  }
+  if (content && content.json) {
+    try {
+      logger.debug(`Emitting JSON from ${typeof content.json}`);
+      return {
+        response: {
+          body: JSON.stringify(content.json),
```
trieloff: I'm not sure about that. I thought OpenWhisk expects JSON responses as object literals. At least in the static action we check for text/binary/json and treat each one differently.

trieloff: I'm not sure about that. I thought OpenWhisk expects JSON responses as object literals. At least in the static action we check for text/binary/json and treat each one differently.

src/defaults/json.pipe.js
rofe: ;)

src/json/emit-json.js
rofe: You are right: local output is getting over-escaped, too 😬 
Should we use `JSON.parse(JSON.stringify(content.json))` in order to be sure the result will be JSON-compatible? Or simply pass `content.json` into the `response.body` (and hope for the best)?

Issue #279, opened 2019-04-23
Default pipeline should include <section/>s


davidnuescheler:
currently the default pipeline renders thematic breaks (`---`) in markdown to `<hr>` and while it is a departure from githubs rendering they probably should correctly translate to `<section> ... </section>` which definitely is a lot more useful for DOM query selectors.

trieloff:
@koraa do you remember changing this?

kptdobe:
It is not a regression, it is an "evolution". The default `document` has always been separated from the `sections` handling": it has contained the "raw" html without extra processing. The ask here is to at least generate the `<section>` tags into the default `document`.

koraa:
@trieloff What do you mean?
@davidnuescheler See https://github.com/adobe/helix-pipeline/pull/260 :)

trieloff:
I don't remember ever seeing `<hr>` in the DOM, so I wondered if this might be a regression, but it looks like the fix (i.e. #260) is what's missing.

kptdobe:
I am almost sure that the code that handles the sections never touched the original document and I have seen the `<hr>`. `sections` have always been a separate concept, if you want to use them, you need to iterate and output each section.
The ask here is that `context.content.document` contains the `<section>` tags instead of the `<hr>` (and the correct children).

trieloff:
Well, then it is an enhancement indeed.

koraa:
@trieloff When I test this on the current pipeline, <hr/> is rendered as just that – hr. I think we just never really did use <hr/> much in our examples :)
AFAICS, the current behaviour is not really a bug – the way sections where supposed to be used until now was just to use the content.sections property and manually convert to html. This is indeed what #260 is intended to change :)

trieloff:
Good. But I agree with @kptdobe that it would make sense to split write-protection removal and section handling into two separate PRs.

kptdobe:
ah ah, I think we all agree then ;)
We just reached the point where we think sections are a good thing and should now be part of the "default".
Note that if there is no thematic break in the document, you will end up with one single section. That's the current behavior, I think it makes sense to keep it but it can be debated. Question might be raised at one point.

kptdobe:
Note also that #143 is a slightly different thing which might be invalidated by the current issue: idea here was to have a VDOM object for each section which will get for free once the sections are in the content.document object. I am going to close #143 and use this one as a replacement.

trieloff:
[...]

PR #301, opened 2019-05-02
fix(sections): show sections in DOM

trieloff:


davidnuescheler:
as mentioned in #279 i think that it is counter intuitive to wrap an `.md` without any `ThematicBreak` into a `<section>`, so i am proposing a small change in the `split-sections.js`

koraa:
@davidnuescheler Unfortunately I have to disagree about not using sections in a document without thematic breaks. Not wrapping those documents would create a big need for special casing and present an issue for the usability of our html output:

With wrapping you could always assume that: `document.body.children` is a list of sections in the htl if you want to get the content of the document; with this special case one would always test whether document.body.contains no section tags and then fall back to `[document.body]`. This seems a little esoteric I know, but I believe this would be a significant downside in many actual htl configuration (we may test this by looking for a few common htl usage patterns and seeing how the special casing affects them).

Another use case that may be a problem is styling with `section { ... }`; you could never assume sections are there.

Is there any actual issue with having a single section in a document? I believe unless some custom css or code uses sections, a section tag would pretty much not change the rendering in any way?

davidnuescheler:
I think I changed my mind and i am following @koraa @trieloff that it is better to special case the usecases where you need want to drop the wrapping section than to remove the sections by default if there is no `ThematicBreak` 

also based on the discussion in #279 i think we should change this to `<div class="section">` instead.

src/utils/section-handler.js
```diff
@@ -0,0 +1,22 @@
...
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+function section() {
+  return function handler(h, node, _, handlechild) {
+    const sectionnode = h(node, 'div', { class: 'section' });
```
ramboz: Maybe `sectioNode`?

Also, should we prefix the class to avoid potential collisions with other frameworks or the customer classes?

What about following https://suitcss.github.io/ and use something like `hlx-section`

kptdobe:
@trieloff Is this still in progress ?

kptdobe:
@trieloff I was more thinking of a merge than a close... ;) Any explanation why you closed it ? We need the change to clean up `helix-pages`.

PR #326, opened 2019-05-10
feat(html pipe): Handling 404 Errors

### LINKED ISSUE #227: Idea: Handling 404 Errors
Here's an idea on handling 404 errors in a Helix-ish way.

Assumptions:
- customers want nice (customizable) 404 error pages
- the HTML pipeline knows that there is a 404 error after the `fetch` step

Solution:

1. after the `fetch` step, add another `error` step, which will do the following
2. output HTML that has following content: `<esi:include src="404.html">`
3. finish the rendering pipeline
4. Fastly will then either include a static `404.html` from the `static` repo, giving you full customizability, or go back to the html pipeline, which is this time retrieving `404.md` from the `content` repository and renders it

Open Questions:

- [ ] how to make sure we don't end up in an endless loop if `404.md` doesn't exist
- [ ] how to make sure the correct status code (404) gets set when serving the static `404.html`

ramboz:
Graceful error handling with custom pages and proper fallback

fix #227

### DIFF
--- a/src/html/fetch-markdown.js
+++ b/src/html/fetch-markdown.js
@@ -11,7 +11,7 @@
  */
 const client = require('request-promise-native');
 const URI = require('uri-js');
-const { bail } = require('../helper');
+const { bail, safebail } = require('../helper');
 
 function uri(root, owner, repo, ref, path) {
   const rootURI = URI.parse(root);
@@ -97,6 +97,10 @@ async function fetch(
     };
   } catch (e) {
     if (e && e.statusCode && e.statusCode === 404) {
+      if (path.match(/^\/?404\.(md|html)$/)) {
+        // Passing status 200 since we'll first have tried 400.html with a 404
+        return safebail(logger, `Could not find Markdown at ${options.uri}`, e, 200);
+      }
       return bail(logger, `Could not find Markdown at ${options.uri}`, e, 404);
     } else if ((e && e.response && e.response.elapsedTime && e.response.elapsedTime > timeout) || (e && e.cause && e.cause.code && (e.cause.code === 'ESOCKETTIMEDOUT' || e.cause.code === 'ETIMEDOUT'))) {
       // return gateway timeout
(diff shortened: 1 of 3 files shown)


src/html/fetch-markdown.js
```diff
@@ -97,6 +97,10 @@ async function fetch(
     };
   } catch (e) {
     if (e && e.statusCode && e.statusCode === 404) {
+      if (path.match(/^\/?404\.(md|html)$/)) {
```
ramboz: The idea is that we will first do a regular `bail` with the proper status code (i.e. `400`, `500`, etc.) and try to fetch the corresponding md file (`400.md`, `500.md`, etc.).
If that is missing, we'll end up here again with a path that matches 404, and perform a `safebail` as fallback

trieloff:
In terms of traffic control, I'd like to merge #260 first, even if we still have to discuss through a couple of issues that would affect the handling of errors (and this PR)

src/html/fetch-markdown.js
tripodsan: hmm, I would do this in an/the error handler at the end of the pipe, and not in fetch.

src/html/fetch-markdown.js
```diff
@@ -97,6 +97,10 @@ async function fetch(
...
+        // Passing status 200 since we'll first have tried 400.html with a 404
+        return safebail(logger, `Could not find Markdown at ${options.uri}`, e, 200);
```
tripodsan: this is IMO wrong: you must not return a `200`. otherwise fastly doesn't know that the resource is missing and cannot fallback to static processing...which actually renders this entire feature useless :-)

trieloff: We can control the fallback, but it should definitely be a 404.

tripodsan:
since we do error handling now in the dispatcher, I'll close this for now.

Issue #337, opened 2019-05-22
use jsdom throughout the pipeline and use jsdom supporting htl engine


tripodsan:
once the htl engine supports direct jsdom injection (see https://github.com/adobe/htlengine/issues/62), it should be use in the pipeline.

rofe:
Once that is the case, can we resurrect https://github.com/adobe/htlengine/issues/52? 😬

tripodsan:
> Once that is the case, can we resurrect [adobe/htlengine#52](https://github.com/adobe/htlengine/issues/52)? 😬

yes. we could parse the injected HTML context with JSDOM and then use the dom-purifier on it. but only if JSDOM is enabled - so we need to keep the other sanitizer as well.

Issue #336, opened 2019-05-22
improve vdom generator performance


tripodsan:
- separate to the toHTAST step completely
- don't reparse HTML generated by the handlers
- avoid toHtml/parseHTML rountripds
- generate jsdom or hast directly in the handlers

tripodsan:
some key measurements with a 23k markdown:

|   | md-2-html | once | reparse | stringify |
|--|------------|------|---------|---------|
| before| 44ms | 23ms| 20s  | 2ms |
| after   | 27ms | 23ms| 32ms | 2ms |

Once the htlengine directly outputs JSDOM, the `reparse` should drop to 0ms.

PR #349, opened 2019-05-29
feat(pipe): use jsdom everywhere

tripodsan:
see #337, see #336 

## Improvements
- remove the use of `htast` and `hast` where possible
- the VDOM transformer now directly transforms the `mdast` to `hast` to `dom`. before it used to create the jsdom via a html-string roundtrip.

## BREAKING CHANGES:
- remove `htast` from `context.content` schema
- remove `hast` from `context.response` schema
- add `document` from to `context.response` schema
- `context.htast` and `context.response.hast` is no longer generated
- mdast to jsdom transformation handlers can only return hast (not string, not vdom)
- html postprocessing pipeline step was renamed from `hast` to `post`
- mixing inline html and markdown no longer works (eg: `<em>This is **strong**</em>`)

## Further Work

I think that we should generate the jsdom directly from the `mdast`. this would even speed up the transformation more. currently, there is no utility available for this. we would need to copy major parts of [mdast-util-to-hast](https://github.com/syntax-tree/mdast-util-to-hast).


src/defaults/html.pipe.js
```diff
@@ -63,18 +63,20 @@ const htmlpipe = (cont, payload, action) => {
     .before(selectstrain)
     .before(selecttest)
     .before(html).expose('html')
-    .before(responsive)
+    // todo: responsive used to operate on the htast, therefore ignored if content.document was used
+    // todo: there is similar logic in the image-handler during jsdom creation....
+    // .before(responsive)
```
tripodsan: @trieloff WDYT?

src/utils/hast-util-to-dom.js
```diff
@@ -0,0 +1,179 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+ * REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ * FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+ * INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS
+ * OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+ * TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+ * THIS SOFTWARE.
+ */
+
+// This was copied from https://github.com/syntax-tree/hast-util-to-dom/blob/master/src/index.js
```
tripodsan: see https://github.com/syntax-tree/hast-util-to-dom/issues/3

ramboz:
@tripodsan These are triggered when you use custom HTML elements and attributes. So Something like:
```
<my-tag has-some-attribute/>
```

I have a test for those in https://github.com/adobe/helix-pipeline/blob/051375678c3a6280e437e89dc693559b25fae2e9/test/testHTMLFromMarkdown.js#L304-L316, so you can probably leverage something similar

tripodsan:
@ramboz when I run your test, it never gets into this function. I don't know what changed...but the test succeeds.

ramboz:
@tripodsan Alright. I'll review this tomorrow

PR #351, opened 2019-05-30
test(markdown): actually fail if dom is not equal

tripodsan:


test/testHTMLFromMarkdown.js
```diff
@@ -296,22 +289,21 @@ describe('Testing Markdown conversion', () => {
         # location
         <a name="anchors">Foo</a>
       `, `
-        <h1 id="location">location</h1>
+        <h1>location</h1>
```
tripodsan: @ramboz I think it indeed should remove the location attribute, right?

ramboz: 🤦‍♂ yes!

test/testHTMLFromMarkdown.js
```diff
@@ -98,33 +96,28 @@ const assertMd = async (md, html) => {
     },
   );
 
-  eq(
```
tripodsan: @ramboz the tests never actually failed, because the `eq` only tests for equality. but doesn't throw if not equal :-)

PR #358, opened 2019-06-06
feat(html-pipe): disable sanitizer by default

### LINKED ISSUE #356: Disable XSS sanitizer by default
Allowing HTML in markdown voids the need to a XSS sanitizer a little bit.
Suggest to disable it by default. either by removing it from the html pipeline completely, or safeguard it with:

```
    .before(sanitize).when(paranoid)
```

/cc @trieloff @davidnuescheler @ramboz

tripodsan:
fixes #356

### DIFF
--- a/src/defaults/html.pipe.js
+++ b/src/defaults/html.pipe.js
@@ -44,6 +44,10 @@ function hascontent({ content }) {
   return !(content !== undefined && content.body !== undefined);
 }
 
+function paranoid(context, action) {
+  return action && action.secrets && !!action.secrets.HLX_SANITIZE_DOM;
+}
+
 const htmlpipe = (cont, context, action) => {
   action.logger = action.logger || log;
   action.logger.log('debug', 'Constructing HTML Pipeline');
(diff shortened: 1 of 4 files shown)


src/defaults/html.pipe.js
```diff
@@ -44,6 +44,10 @@ function hascontent({ content }) {
   return !(content !== undefined && content.body !== undefined);
 }
 
+function paranoid() {
+  return !!process.env.HLX_SANITIZE_DOM;
```
trieloff: We should pass the `paranoid` flag through params, not through environment vars, as the latter can't be set easily on a function-by-function basis.

tripodsan: it could be set in the pre.js :-)
```
if (action.params.paranoid) {
  process.env.HLX_SANITIZE_DOM = true;
}
```

but I'll add it to the `secrets`

trieloff: I think each invocation is running in a separate process, but it's not a good idea anyway.

src/defaults/html.pipe.js
```diff
@@ -44,6 +44,10 @@ function hascontent({ content }) {
   return !(content !== undefined && content.body !== undefined);
 }
 
+function paranoid(context, action) {
+  return action && action.secrets && !!action.secrets.HLX_SANITIZE_DOM;
```
rofe: @trieloff shouldn't `HLX_SANITIZE_DOM` also be added to https://github.com/adobe/helix-pipeline/blob/master/src/schemas/secrets.schema.json?

PR #367, opened 2019-06-10
feat(pipe): keep context dumps in memory

### LINKED ISSUE #331: context dump gets really large and adds a lot of overhead
for larger requests, the context dumps of 1 request can get large ( > 20mb). we should add a possibility to disable or avoid large dumps.

tripodsan:
fixes #331

### DIFF
--- a/src/utils/dump-context.js
+++ b/src/utils/dump-context.js
@@ -12,9 +12,24 @@
 
 const path = require('path');
 const fs = require('fs-extra');
+const { setdefault } = require('@adobe/helix-shared').types;
 
-function tstamp() {
-  const now = new Date();
+
+/**
+ * Returns {@code true} if context dumps should never be written to disk.
+ * @returns {boolean} {@code true} if writing dumps is disabled.
+ */
+function disableDisk() {
+  // if we run in openwhisk, we never want files to be written to disk.
+  return !!process.env.__OW_ACTIVATION_ID;
+}
+
+/**
+ * Creates a timestamp string of the given date.
+ * @param {Date} now the time
+ * @returns {string} a timestamp string
+ */
+function tstamp(now) {
   let retstr = '';
   retstr += (`${now.getFullYear()}`).padStart(4, '0');
   retstr += (`${now.getUTCMonth()}`).padStart(2, '0');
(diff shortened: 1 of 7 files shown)


src/utils/dump-context.js
```diff
@@ -32,20 +47,78 @@ function tstamp() {
...
+  const debug = setdefault(action, 'debug', { contextDumps: [] });
+  const info = {
+    index,
+    name: name ? name.split(':').pop() : 'undef',
+    time: new Date(),
+    json: JSON.stringify(context, null, 2),
+  };
+  debug.contextDumps.push(info);
+
+  if (action.logger.level === 'silly' && !disableDisk()) {
```
kptdobe: From the issue:

> always dump it for log-level debug and silly

Only `silly` seems to have been implemented.

src/utils/dump-context.js
```diff
@@ -32,20 +47,78 @@ function tstamp() {
...
-  const dumppath = path.resolve(action.debug.dumpDir, `context-${nowStr}-step-${index}.json`);
-  await fs.writeJson(dumppath, context, { spaces: 2 });
-  return dumppath;
+/**
+ * Writes the collected context dumps to disk
+ * @param {*} context Pipeline context
+ * @param {*} action Pipeline action
+ */
+async function report(context, action) {
+  if (action.logger.level === 'silly' || disableDisk()) {
```
kptdobe: Same here

src/utils/dump-context.js
tripodsan: yeah, I though that maybe with `debug` we pollute too much. from my own experience, I often use `debug`, but then I guess, I don't need to dumps very often.....

but I'm indifferent.

kptdobe: I am fine with `silly` only, just wanted to mention it...

src/utils/dump-context.js
```diff
@@ -12,9 +12,24 @@
...
+const { setdefault } = require('@adobe/helix-shared').types;
 
-function tstamp() {
-  const now = new Date();
+
+/**
+ * Returns {@code true} if context dumps should never be written to disk.
+ * @returns {boolean} {@code true} if writing dumps is disabled.
+ */
+function disableDisk() {
```
trieloff: I'd use the `production` check we already have, which guards against exceptions when accessing the environment vars directly.

src/utils/dump-context.js
```diff
@@ -32,20 +47,78 @@ function tstamp() {
...
+  if (action.debug.dumpDir) {
+    return;
+  }
+  const stamp = tstamp(new Date());
+  const id = (context.request && context.request.headers && context.request.headers['x-openwhisk-activation-id']) || '';
+  const pathStr = context.request && context.request.path ? context.request.path.replace(/\//g, '-').substring(1) : '';
+  const dirName = `context_dump_${stamp}_${pathStr}_${id}`;
+  action.debug.dumpDir = path.resolve(process.cwd(), 'logs', 'debug', dirName);
+  await fs.ensureDir(action.debug.dumpDir);
+  action.logger.info(`Writing context dumps to: ${action.debug.dumpDir}`);
```
trieloff: should be `debug` level

src/utils/dump-context.js
trieloff: `silly` is fine.

src/utils/dump-context.js
tripodsan: - I think `production` is the wrong test for this purpose. I just don't want to dump in openwhisk - no matter if it's production or stage or dev...etc....
- `!!process.env.*` will never thrown an exception (because `process.env` is guaranteed to exist in nodejs)

src/utils/dump-context.js
tripodsan: no. in case it dumps because of an error, do I want to see that information in the log right away.

src/utils/dump-context.js
stefan-guggisberg: > `!!process.env.*` will never thrown an exception (because `process.env` is guaranteed to exist in nodejs)

you would assume so, yes. but:
https://github.com/adobe/helix-pipeline/issues/142
https://github.com/adobe/helix-cli/issues/408#issuecomment-448621464

tripodsan: hmmm.... I think this was a temporary problem....

see my test:
```js
function main(params) {
  return {
    statusCode: 200,
    headers: { 'Content-Type': 'text/html', 'X-Foo': 'bar' },
    body: JSON.stringify(process.env, null, 2),
  };
}
```


```
$ curl https://adobeioruntime.net/api/v1/web/helix-pages/default/test
{
  "__OW_ACTION_NAME": "/helix-pages/test",
  "__OW_ACTIVATION_ID": "9f4a78a9ed2a40818a78a9ed2a60810d",
  "__OW_DEADLINE": "1560327247499",
  "__OW_NAMESPACE": "helix-pages",
  "NODE_VERSION": "10.15.1",
  "HOSTNAME": "a6d134123987",
  "YARN_VERSION": "1.13.0",
  "SHLVL": "1",
  "HOME": "/home/runtimeuser",
  "PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
  "NOT_ROOT_USER": "runtimeuser",
  "MESOS_SANDBOX": "/mnt/mesos/sandbox",
  "__OW_ALLOW_CONCURRENT": "true",
  "UID": "1001",
  "__OW_API_HOST": "https://runtime.adobe.io:443",
  "__OW_JSON_LOG": "true",
  "PWD": "/nodejsAction",
}
```

stefan-guggisberg: I looked closer at the stack trace. The error was thrown by code in the openwhisk node runtime:

```javascript
TypeError: Cannot read property '__OW_ACTIVATION_ID' of null
    at Object.get [as __OW_ACTIVATION_ID] (/nodejsAction/src/service.js:251:60)
    at production (/nodejsAction/XXDPCCdJ/node_modules/@adobe/helix-pipeline/src/utils/is-production.js:18:50)
...
``` 
The node runtime injects dynamic properties into `process.env` in order to enable concurrent activations. There must have been a (temporary?) issue.

tripodsan: they probably have a special getter for the `process.env`, in order to isolate it for concurrent invocations. maybe it was some (scary) raise-condition... 😨

PR #379, opened 2019-06-15
feat(html pipe): Default pipeline should include <section/>s

### LINKED ISSUE #279: Default pipeline should include <section/>s

ramboz:
Include proper sections in the HTML pipeline instead of plain `<hr>` sectioning

BREAKING CHANGE:
- Remove `context.content.sections` and move the sections directly under
`context.content.mdast`
- Adjust json schemas accordingly
- Move the section yaml meta info to `<node>.meta`
- Move the section types to the mdast `<node>.meta.types`

Is dependent on adobe/helix-cli#1037

fix #279

### DIFF
--- a/docs/content.schema.json
+++ b/docs/content.schema.json
@@ -46,7 +46,7 @@
             "description": "The XML object to emit. See xmlbuilder-js for syntax."
         },
         "meta": {
-            "$ref": "https://ns.adobe.com/helix/pipeline/meta#/definitions/meta/properties/meta"
+            "$ref": "https://ns.adobe.com/helix/pipeline/meta#/definitions/meta"
         },
         "title": {
             "$ref": "https://ns.adobe.com/helix/pipeline/meta#/definitions/meta/properties/title"
--- a/docs/content.schema.json
+++ b/docs/content.schema.json
@@ -33,13 +33,6 @@
         "mdast": {
             "$ref": "https://ns.adobe.com/helix/pipeline/mdast"
         },
-        "sections": {
-            "type": "array",
-            "description": "The extracted sections of the document",
-            "items": {
-                "$ref": "https://ns.adobe.com/helix/pipeline/section"
-            }
-        },
         "document": {
             "type": "object",
             "description": "The DOM-compatible representation of the document's inner HTML"
--- a/docs/mdast.schema.md
+++ b/docs/mdast.schema.md
@@ -348,10 +348,9 @@ For code, a lang field can be present. It represents the language of computer co
 | `class`| string | Optional |
 | `image`| string | Optional |
 | `intro`| string | Optional |
-| `meta`| boolean | Optional |
 | `tagname`| string | Optional |
 | `title`| string | Optional |
-| `types`| boolean | Optional |
+| `types`| array | Optional |
 
 
 
@@ -422,26 +421,6 @@ Extracted first paragraph of the document
 
 
 
-#### meta
-
-When set to `true`, will output the additional meta properties as data attributes (i.e. `'foo = bar'` will become `data-hlx-foo='bar'`)
-
-`meta`
-
-* is optional
-* type: `boolean`
-
-##### meta Type
-
-
-`boolean`
-
-
-
-
-
-
-
 #### tagname
 
 The element tag name to use for the section instead of the default `div` one (i.e. `section`, `main`, `aside`)
@@ -488,17 +467,26 @@ Extracted title of the document
 
 #### types
 
-When set to `true`, will output the section types as additional CSS classes (i.e. `class='hlx-section has-image has-heading'`)
+The inferred class names for the section
 
 `types`
 
 * is optional
-* type: `boolean`
+* type: `string[]`
+
 
 ##### types Type
 
 
-`boolean`
+Array type: `string[]`
+
+All items must be of the type:
+`string`
+
+
+
+
+
 
 
 
--- a/docs/mdast.schema.md
+++ b/docs/mdast.schema.md
@@ -27,9 +27,11 @@ A node in the Markdown AST
 | [data](#data) | `object` | Optional  | No | MDAST (this schema) |
 | [depth](#depth) | `integer` | Optional  | No | MDAST (this schema) |
 | [identifier](#identifier) | `string` | Optional  | No | MDAST (this schema) |
+| [image](#image) | `string` | Optional  | Yes | MDAST (this schema) |
+| [intro](#intro) | `string` | Optional  | Yes | MDAST (this schema) |
 | [label](#label) | `string` | Optional  | No | MDAST (this schema) |
 | [lang](#lang) | `string` | Optional  | Yes | MDAST (this schema) |
-| [meta](#meta) | `string` | Optional  | Yes | MDAST (this schema) |
+| [meta](#meta) | `object` | Optional  | Yes | MDAST (this schema) |
 | [ordered](#ordered) | `boolean` | Optional  | No | MDAST (this schema) |
 | [payload](#payload) | `object` | Optional  | No | MDAST (this schema) |
 | [position](#position) | Position | Optional  | No | MDAST (this schema) |
@@ -38,6 +40,7 @@ A node in the Markdown AST
 | [start](#start) | `integer` | Optional  | Yes | MDAST (this schema) |
 | [title](#title) | `string` | Optional  | Yes | MDAST (this schema) |
 | [type](#type) | `enum` | Optional  | No | MDAST (this schema) |
+| [types](#types) | `string[]` | Optional  | No | MDAST (this schema) |
 | [url](#url) | `string` | Optional  | No | MDAST (this schema) |
 | [value](#value) | `string` | Optional  | No | MDAST (this schema) |
 
@@ -147,6 +150,26 @@ A checked field can be present. It represents whether the item is done (when tru
 
 Array type: `array`
 
+All items must be of the type:
+
+**One** of the following *conditions* need to be fulfilled.
+
+
+#### Condition 1
+
+
+* []() – `https://ns.adobe.com/helix/pipeline/mdast`
+
+
+#### Condition 2
+
+
+* []() – `https://ns.adobe.com/helix/pipeline/section#/definitions/section`
+
+
+  
+
+
 
 
 
@@ -219,6 +242,49 @@ For associations, an identifier field must be present. It can match an identifie
 
 
 
+## image
+
+Path (can be relative) to the first image in the document
+
+`image`
+
+* is optional
+* type: `string`
+* defined in this schema
+
+### image Type
+
+
+`string`, nullable
+
+* format: `uri-reference` – URI Reference (according to [RFC3986](https://tools.ietf.org/html/rfc3986))
+
+
+
+
+
+
+## intro
+
+Extracted first paragraph of the document
+
+`intro`
+
+* is optional
+* type: `string`
+* defined in this schema
+
+### intro Type
+
+
+`string`, nullable
+
+
+
+
+
+
+
 ## label
 
 For associations, a label field can be present. It represents the original value of the normalised identifier field.
@@ -263,19 +329,22 @@ For code, a lang field can be present. It represents the language of computer co
 
 ## meta
 
-For code, if lang is present, a meta field can be present. It represents custom information relating to the node.
+Extracted metadata from the frontmatter of the document
 
 `meta`
 
 * is optional
-* type: `string`
+* type: `object`
 * defined in this schema
 
 ### meta Type
 
 
-`string`, nullable
+`object`, nullable, with following properties:
 
+
+| Property | Type | Required |
+|----------|------|----------|
 
 
 
@@ -407,7 +476,7 @@ Starting item of the list
 
 ## title
 
-For resources, a title field can be present. It represents advisory information for the resource, such as would be appropriate for a tooltip.
+Extracted title of the document
 
 `title`
 
@@ -470,6 +539,33 @@ The value of this property **must** be equal to one of the [known values below](
 | `embed` | Content embedded from another page, identified by the `url` attribute. |
 | `section` | A section within the document. Sections serve as a high-level structure of a single markdown document and can have their own section-specific front matter metadata. |
 | `listItem` |  |
+
+
+
+
+## types
+
+The inferred class names for the section
+
+`types`
+
+* is optional
+* type: `string[]`
+* defined in this schema
+
+### types Type
+
+
+Array type: `string[]`
+
+All items must be of the type:
+`string`
+
+
+
+
+
+
 
 
 
--- a/docs/mdast.schema.md
+++ b/docs/mdast.schema.md
@@ -23,7 +23,7 @@ A node in the Markdown AST
 | [align](#align) | `enum[]` | Optional  | No | MDAST (this schema) |
 | [alt](#alt) | `string` | Optional  | Yes | MDAST (this schema) |
 | [checked](#checked) | `boolean` | Optional  | Yes | MDAST (this schema) |
-| [children](#children) | MDAST | Optional  | No | MDAST (this schema) |
+| [children](#children) | `array` | Optional  | No | MDAST (this schema) |
 | [data](#data) | `object` | Optional  | No | MDAST (this schema) |
 | [depth](#depth) | `integer` | Optional  | No | MDAST (this schema) |
 | [identifier](#identifier) | `string` | Optional  | No | MDAST (this schema) |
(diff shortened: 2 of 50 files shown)


ramboz:
Trying to gather some early feedback to validate the approach before spending too much time on re-writing the tests

src/html/get-metadata.js
```diff
@@ -19,7 +19,7 @@ const {
 
 function yaml(section) {
   const yamls = selectAll('yaml', section);
-  section.meta = obj(flat(map(yamls, ({ payload }) => payload)));
+  section.meta = Object.assign({}, obj(flat(map(yamls, ({ payload }) => payload))));
```
trieloff: Why the additional `Object.assign`?

src/html/split-sections.js
```diff
@@ -11,8 +11,15 @@
  */
 
 const between = require('unist-util-find-all-between');
+const { selectAll } = require('unist-util-select');
+const { flat, obj, map } = require('@adobe/helix-shared').sequence;
```
trieloff: I'd import `@adobe/ferrum` directly

src/html/get-metadata.js
```diff
@@ -144,7 +144,8 @@ function fallback(section) {
 }
 
 function getmetadata({ content }, { logger }) {
-  const { sections } = content;
+  const { mdast: { children } } = content;
+  const sections = children.filter(node => node.type === 'root');
```
trieloff: Can this work? Below, you are changing the type to `section`.

src/html/split-sections.js
```diff
@@ -25,18 +32,25 @@ function split({ content }) {
...
   // but filter out empty section
     .filter(([start, end]) => start !== end)
   // then return all nodes that are in between
     .map(([start, end]) => {
     // skip 'thematicBreak' nodes
       const index = mdast.children[start].type === 'thematicBreak' ? start + 1 : start;
-      return {
-        type: 'root',
+      const section = {
+        type: 'section',
```
trieloff: New node type is a good idea.

src/schemas/mdast.schema.json
```diff
@@ -76,15 +77,21 @@
         "imageReference": "A pointer to an image",
         "footnote": "A footnote",
         "footnoteReference": "A reference to a footnote",
-        "embed": "Content embedded from another page, identified by the `url` attribute."
+        "embed": "Content embedded from another page, identified by the `url` attribute.",
+        "section": "The extracted sections of the document"
```
trieloff:
```suggestion
        "section": "A section within the document. Sections serve as a high-level structure of a single markdown document and can have their own section-specific front matter metadata."
```

src/utils/section-handler.js
```diff
@@ -0,0 +1,65 @@
...
+const wrap = require('mdast-util-to-hast/lib/wrap');
+
+/**
+ * Get the tag name for the specified section.
+ *
+ * @param {Node} section The MDAST section to get the tag name for
+ * @returns {string} The tag name for the section. Defaults to {@code div}.
+ */
+function getTageName(section) {
+  return (section.meta && section.meta.tagName) || 'div';
```
trieloff: Can we not use camelCase for YAML properties? `tagname` should be sufficient.

src/utils/section-handler.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/**
+ * Get the class name for the specified section.
+ *
+ * @param {Node} section The MDAST section to get the class name for
+ * @returns {string} The class name for the section. Defaults to {@code hlx-section}.
+ */
+function getClass(section) {
+  return (section.meta && section.meta.className) || 'hlx-section';
```
trieloff: Same here, the YAML property should be `class`, even if it is slightly inconvenient to read it from JavaScript.

Also, what happens to the detected section types?

src/utils/section-handler.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function sectionHandler() {
+  return function handler(h, node, parent) {
+    const n = Object.assign({}, node);
+
+    // we have a section that is not the document root, so wrap it in the desired tag
+    if (parent && parent.type === 'root' && parent.children.length > 1) {
+      const tagName = getTageName(n);
+      const props = { class: getClass(n), 'data-hlx-types': getTypes(n) };
```
trieloff: I'd rather merge the `types` into `class` and then have `data-<prop` for each additional `prop` that's in the section metadata.

src/utils/section-handler.js
```diff
@@ -0,0 +1,65 @@
...
+const all = require('mdast-util-to-hast/lib/all');
+const wrap = require('mdast-util-to-hast/lib/wrap');
+
+/**
+ * Get the tag name for the specified section.
+ *
+ * @param {Node} section The MDAST section to get the tag name for
+ * @returns {string} The tag name for the section. Defaults to {@code div}.
+ */
+function getTageName(section) {
```
rofe: typo: `getTagName`

src/utils/section-handler.js
```diff
@@ -0,0 +1,65 @@
...
+  return types ? types.join(' ') : null;
+}
+
+function sectionHandler() {
+  return function handler(h, node, parent) {
+    const n = Object.assign({}, node);
+
+    // we have a section that is not the document root, so wrap it in the desired tag
+    if (parent && parent.type === 'root' && parent.children.length > 1) {
+      const tagName = getTageName(n);
```
rofe: typo: `getTagName`

src/html/get-metadata.js
ramboz: 👍 Some left-overs from testing various approaches. I'll definitely remove in the final PR.

src/html/get-metadata.js
ramboz: 👍 Again, some left-overs from playing around… though it should have failed somehow. Will double-check

src/html/split-sections.js
ramboz: Is it ok to just inline the node type or do we have a central place to expose them as enum constants that I could use?

src/utils/section-handler.js
ramboz: @trieloff So, something like this?
```
<div class="hlx-Section hlx-Section--hasImage hlx-Section--hasText" data-hlx-<meta1>="<value1>" data-hlx-<meta2>="<value2>" …></div>
```

And we let the dev:
- overwrite the section class
- toggle whether type classes are added (and use the custom section class as prefix if set)
- toggle whether additional meta props are exposed as data attributes

ramboz: @trieloff So, something like this?
```
<div class="hlx-Section hlx-Section--hasImage hlx-Section--hasText" data-hlx-<meta1>="<value1>" data-hlx-<meta2>="<value2>" …></div>
```

And we let the dev:
- overwrite the section class
- toggle whether type classes are added (and use the custom section class as prefix if set)
- toggle whether additional meta props are exposed as data attributes

trieloff: Yes, except that classnames should be lowercase and should be kept as they are (your lower example).

We can fix the collisions later, i.e. through a step that performs class renaming.

My reason for this is that the HTML is simpler, and writing CSS selectors for attributes is harder than for classes, so I'd err on the side of simplicity over consistency.

ramboz:
@trieloff Mostly the tests are failing since I did not touch them yet. Wanted to first validate the approach

src/utils/section-handler.js
ramboz: @trieloff If we assume a step that does the renaming, then does it make sense to let the author specify the section class name in the first place in the yaml? I could just hardcode `hlx-section` then.

Less logic => fewer risk of bugs

trieloff:
Keep going, you are on the right track.

docs/content.schema.json
```diff
@@ -33,13 +33,6 @@
         "mdast": {
             "$ref": "https://ns.adobe.com/helix/pipeline/mdast"
         },
-        "sections": {
```
ramboz: BREAKING CHANGE: Sections is removed and added to the `mdast` directly

docs/mdast.schema.md
```diff
@@ -23,13 +23,15 @@ A node in the Markdown AST
...
+| [children](#children) | `array` | Optional  | No | MDAST (this schema) |
 | [data](#data) | `object` | Optional  | No | MDAST (this schema) |
 | [depth](#depth) | `integer` | Optional  | No | MDAST (this schema) |
 | [identifier](#identifier) | `string` | Optional  | No | MDAST (this schema) |
+| [image](#image) | `string` | Optional  | Yes | MDAST (this schema) |
+| [intro](#intro) | `string` | Optional  | Yes | MDAST (this schema) |
 | [label](#label) | `string` | Optional  | No | MDAST (this schema) |
 | [lang](#lang) | `string` | Optional  | Yes | MDAST (this schema) |
-| [meta](#meta) | `string` | Optional  | Yes | MDAST (this schema) |
+| [meta](#meta) | `object` | Optional  | Yes | MDAST (this schema) |
```
ramboz: BREAKING CHANGE: the `meta` now contains the `yaml` meta information

src/schemas/content.schema.json
```diff
@@ -33,13 +33,6 @@
     "mdast": {
       "$ref": "https://ns.adobe.com/helix/pipeline/mdast"
     },
-    "sections": {
```
ramboz: BREAKING CHANGE: as above in the doc

src/schemas/mdast.schema.json
```diff
@@ -148,9 +153,29 @@
       "format": "uri-reference",
       "description": "For resources, an url field must be present. It represents a URL to the referenced resource."
     },
+    "meta": {
+      "type": ["null", "object"],
```
ramboz: BREAKING CHANGE: as above in the doc

src/utils/section-handler.js
```diff
@@ -0,0 +1,83 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+const all = require('mdast-util-to-hast/lib/all');
+const wrap = require('mdast-util-to-hast/lib/wrap');
+
+const DEFAULT_SECTION_TAG = 'div';
+const DEFAULT_SECTION_CLASS = 'hlx-section';
+const SYSTEM_META_PROPERTIES = ['class', 'meta', 'tagname', 'types'];
```
ramboz: YAML properties:
- `class {string}`: will allow overwriting the default section class (i.e. `hlx-section`)
- `meta {boolean}`: when set to `true` will output any additional YAML properties as `data-hlx-*` attributes
- `tagname {string}`: will allow overwriting the default section tag (i.e. `div`)
- `types {boolean}`: when set to `true` will output the section types as additional classes on the section (i.e. `has-image`, etc.)

test/fixtures/sections/2images.json
```diff
@@ -1,42 +1,40 @@
-[
```
ramboz: Just unwrapping the array since we unwrap sole sections directly on the mdast

test/fixtures/sections/complex.json
```diff
@@ -1,2356 +1,2359 @@
...
-          }
-        ],
-        "data": {
-          "types": [
-            "is-text"
-          ]
-        }
+{
+  "type": "root",
+  "children": [
```
ramboz: Just wrapping the array inside the mdast root's children

test/fixtures/sections/headerimage.json
```diff
@@ -1,50 +1,48 @@
-[
```
ramboz: Just unwrapping the array since we unwrap sole sections directly on the mdast

(same for rest of the sections fixtures below)

test/testConditionalSections.js
```diff
@@ -441,7 +453,7 @@ Or that one at the same time, because they are both part of an A/B test.
     // format, section metadata, the mdast format or the input above, you probably
     // want to change one of these values until you find one that *happens* to work
     // again...
-    assert.notDeepEqual(await runpipe('foo'), await runpipe('bang'));
+    assert.notDeepEqual(await runpipe('foo'), await runpipe('bar'));
```
ramboz: As per existing comment above, the test started failing after my changes, so tried out some other value

test/testGetMetadata.js
```diff
@@ -29,7 +29,7 @@ function callback(body) {
   parseFront(dat);
   split(dat, { logger });
   getmetadata(dat, { logger });
-  return dat.content.sections;
+  return dat.content.mdast;
```
ramboz: Expose the `mdast` node directly since most tests have a single section that will be unwrapped to it

test/testGetMetadata.js
```diff
@@ -55,26 +55,7 @@ describe('Test getMetadata', () => {
...
-          position: {},
-          type: '',
-        },
-      },
-    };
-    getmetadata(dat, { logger });
-    assert.deepEqual(dat.content.meta, {});
-  });
-
-  it('getmetadata does not fail with missing sections', () => {
```
ramboz: Removing this test since the empty mdast and missing sections are now the same.

src/html/get-metadata.js
```diff
@@ -158,9 +157,9 @@ function getmetadata({ content }, { logger }) {
 
   const img = sections.filter(section => section.image)[0];
 
-  content.meta = empty(sections) ? {} : sections[0].meta;
```
ramboz: sections will never be empty as we fall back to the `mdast` node

src/utils/section-handler.js
ramboz: Since `meta` and `types` are actual feature flags, should we follow a specific naming convention?
`enable_meta`/`enable_types`, `output_meta`/`output_types`, etc.

trieloff: I would only do that once we expose it as a config parameter.

src/utils/section-handler.js
trieloff: I think allowing to specify the class name is a useful quick fix for some scenarios.

trieloff:
@ramboz for an example of `class` look here: https://github.com/davidnuescheler/character-landing/blob/master/index.md

src/utils/section-handler.js
davidnuescheler: i don't think we need to prefix the `section` with `hlx-section` as i think that the entire default DOM is a helix DOM and in case of a conflict the or some other desire to change the DOM it is very easily done in a the `pre.js` via jquery or regular DOM manipulation.

davidnuescheler:
nice find @trieloff , of course that one didn't work yet. 

i am really looking forward to having this implemented, looks great to me.

ramboz:
After a quite lengthy discussion with @davidnuescheler, the conclusion is:

1. use a plain `<div/>` to wrap the section content. can be overridden with the `tagname` YAML property
2. use a default `hlx-section` class, but custom class can be added via YAML property
3. all valid HTML attributes in a YAML block are added as-is on the section element. Non-valid attributes are set as `data-<attribute>` instead.
4. Section types (i.e. `has-image is-image-only …`) are moved back to the MDAST `meta` object (instead of directly on the meta). And exposed in the resulting DOM as `data-hlx-types` (edited) 
5. Pipeline `after` step removes all `hlx-*` classes and `data-hlx-*` attributes in production code

ramboz:
@trieloff As this requires changes in the IT tests in the cli, I'm not sure how to link the 2 branches together so the tests are passing. Any suggestion?

trieloff:
@ramboz proceed carefully. Merge this branch, knowing that the smoke tests fail (and make sure they only fail in places you expect) then merge the fix in CLI.

ramboz:
@trieloff Sounds good. Would have been great if the smoke test was smart enough to run from the same named branch instead of master in that case.

Anyway, haven't merged directly here before. Anything special in the procedure, or can I just go ahead and squash-merge, then update cli accordingly?

tripodsan:
> just go ahead and squash-merge, then update cli accordingly?

yes.

ramboz:
Include proper sections in the HTML pipeline instead of plain `<hr>` sectioning

BREAKING CHANGE:
- Remove `context.content.sections` and move the sections directly under
`context.content.mdast`
- Adjust json schemas accordingly
- Move the section yaml meta info to `<node>.meta`
- Move the section types to the mdast `<node>.meta.types`

Is dependent on adobe/helix-cli#1037

fix #279

src/html/split-sections.js
rofe: maybe a schema would be in order?

ramboz: @rofe I added a schema, and updated existing ones accordingly with the changes.
Was just wondering if we have a sort of global "constants" file where I would be pulling the string from, or if it is ok to just inline for now

PR #384, opened 2019-06-20
feat(fetch): resolve github ref to commit sha

### LINKED ISSUE #382: Use resolved head ref for content loading
see https://github.com/adobe/helix-experimental-dispatch/issues/3

During fetch, all refs that are not SHAs should be resolved using https://github.com/adobe/helix-resolve-git-ref in order to get a stable github raw url.

tripodsan:
fixes #382

## Question
I currently set the the API endpoint in the secrets schema to the empty string.

```
    "RESOLVE_GITREF_SERVICE": {
      "type": "string",
      "description": "API endpoint to micro service that resolves github refs to commit SHAs.",
      "default": ""
    }
```

This has the advantage, that all action don't use the new lookup right away. so for it to work, we also need to set the action param during `hlx deploy`. This will be required anyways, because we want to execute the lookup action in the users namespace and not the helix one.

### DIFF
--- a/src/defaults/html.pipe.js
+++ b/src/defaults/html.pipe.js
@@ -37,6 +37,7 @@ const tohtml = require('../html/stringify-response');
 const addHeaders = require('../html/add-headers');
 const timing = require('../utils/timing');
 const sanitize = require('../html/sanitize');
+const resolveRef = require('../utils/resolve-ref');
 
 /* eslint newline-per-chained-call: off */
 
@@ -57,6 +58,7 @@ const htmlpipe = (cont, context, action) => {
     .every(dump.record)
     .every(validate).when(() => !production())
     .every(timer.update)
+    .before(resolveRef).expose('resolve').when(hascontent)
     .before(fetch).expose('fetch').when(hascontent)
     .before(parse).expose('parse')
     .before(parseFrontmatter)
--- a/src/schemas/secrets.schema.json
+++ b/src/schemas/secrets.schema.json
@@ -73,6 +73,11 @@
       "type":"boolean",
       "description": "Sanitize the HTML output to guard against XSS attacks. \n\n**Note:** this flag applies a pretty aggressive DOM filtering that will strip out a lot of HTML that your authors might find useful. The setting is meant for processing truly untrusted inputs, such as comments in a social media site.",
       "default":false
+    },
+    "RESOLVE_GITREF_SERVICE": {
+      "type": "string",
+      "description": "API endpoint or action name to the service that resolves github refs to commit SHAs.",
+      "default": ""
     }
   }
 }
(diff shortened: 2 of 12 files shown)


src/utils/resolve-ref.js
```diff
@@ -0,0 +1,83 @@
...
+    ref = 'master';
+  }
+
+  if (REGEXP_SHA.test(ref)) {
+    logger.debug(`github-ref '${ref}' looks like a SHA. Ignoring resolving.`);
+    return;
+  }
+
+  try {
+    const result = await rp({
```
trieloff: I'd use the `openwhisk.action.invoke` function instead of `rp`

src/defaults/html.pipe.js
```diff
@@ -57,6 +58,7 @@ const htmlpipe = (cont, context, action) => {
     .every(dump.record)
     .every(validate).when(() => !production())
     .every(timer.update)
+    .before(resolveRef).expose('resolve').when(hascontent)
```
trieloff: Also add `.when(() => !production())` so that we don't have to emulate on localhost.

tripodsan: no, if the content repo is non-localhost, I also want fresh content...

src/utils/resolve-ref.js
tripodsan: I'd have it rather independent of openwhisk...or do you think because of performance?

src/defaults/html.pipe.js
trieloff: So then we should check for `localhost`

tripodsan: it's done here:
https://github.com/adobe/helix-pipeline/pull/384/files#diff-af7e4527a3d340abcce3d521690db877R35

src/utils/resolve-ref.js
tripodsan: Also, maybe we should keep the naming convention in sync with the embed service, and name it `RESOLVE_GITREF_SERVICE` ?

src/schemas/secrets.schema.json
```diff
@@ -73,6 +73,11 @@
       "type":"boolean",
       "description": "Sanitize the HTML output to guard against XSS attacks. \n\n**Note:** this flag applies a pretty aggressive DOM filtering that will strip out a lot of HTML that your authors might find useful. The setting is meant for processing truly untrusted inputs, such as comments in a social media site.",
       "default":false
+    },
+    "API_RESOLVE_GIT_REF": {
```
tripodsan: @trieloff should we name this `RESOLVE_GITREF_SERVICE` analog to the `EMBED_SERVICE` ?

trieloff: Good idea.

src/utils/resolve-ref.js
trieloff: I think the OpenWhisk performance is better (and OpenWhisk can apply tweaks like running both actions in the same container, etc). I can live with having both, but think that's too much flexibility too early.

PR #391, opened 2019-06-22
fix(all pipe): reduce schema validation error reporting

### LINKED ISSUE #92: Schema validation - disable for now
The schema validation is a really annoying feature : for every single new non-handled case, the rendering the broken. 

New one: 
```
[hlx] warn: tapping failed: Error: Invalid Context at step 2
data.content.mdast.children[9].meta should be string undefined, data.content.mdast.children[11].meta should be string undefined, data.content.mdast.children[13].meta should be string undefined, data.content.mdast.children[15].meta should be string undefined
```
Multiple issues:

1. The logs are `warn` while the validation blocks the rendering. They should be `error`.
2. Appart from adding a break point at the right place in the pipeline (for the record, this line https://github.com/adobe/hypermedia-pipeline/blob/master/src/utils/validate.js#L46), impossible to know what is wrong
3. To unblock the rendering, you need to patch the current hlx version

My proposal here is to keep the messages as `warnings` and to disable the "blocking" aspect of the validation. 

cc @trieloff

ramboz:
fix #92

### DIFF
--- a/src/defaults/default.js
+++ b/src/defaults/default.js
@@ -44,10 +44,15 @@ const log = winston.createLogger({
   ],
 });
 
+function isErrorStep(stepName) {
+  return stepName && stepName.split(':')[0] === 'error';
+}
+
 const defaults = {
   pipe,
   pre,
   log,
+  isErrorStep,
 };
 
 module.exports = defaults;
--- a/src/defaults/default.js
+++ b/src/defaults/default.js
@@ -44,15 +44,10 @@ const log = winston.createLogger({
   ],
 });
 
-function isErrorStep(stepName) {
-  return stepName && stepName.split(':')[0] === 'error';
-}
-
 const defaults = {
   pipe,
   pre,
   log,
-  isErrorStep,
 };
 
 module.exports = defaults;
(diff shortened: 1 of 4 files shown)


trieloff:
Ha. I haven't seen @tripodsan's comment, but it looks like we think alike.

tripodsan:
@ramboz do you plan on working on this?

ramboz:
@tripodsan Somehow missed the comments it seems. I'll pick it up tomorrow

ramboz:
@trieloff, @tripodsan: didn't know about that property. Definitely cleaner!

trieloff:
Not sure why the smoke tests are failing: https://circleci.com/gh/adobe/helix-pipeline/3496?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

ramboz:
Not sure either…
Difference is that I get this in the markup:
```html
<div class="embed  embed-has-title">
  <span class="title">Fastly error: unknown domain embeds.project-helix.io</span>
<div>
```

Could this be a flaky failure just at that time?

trieloff:
I think it's fine to update the smoke tests and then merge.

PR #403, opened 2019-07-08
chore: Migrate to ferrum

koraa:


koraa:
Will merge tomorrow around noon CEST.

stefan-guggisberg:
> Will merge tomorrow around noon CEST.

Please don't merge unless reviewed by @trieloff 

There's also 1 test failing.

test/testHTMLFromMarkdown.js
```diff
@@ -101,9 +100,7 @@ const assertMd = async (md, html, secrets = {}) => {
   // check equality of the dom, but throw assertion based on strings to visualize difference.
   const act = new JSDOM(generated.response.body);
   const exp = new JSDOM(html);
-  if (!eq(act.window.document.body, exp.window.document.body)) {
```
tripodsan: why this change? please keep it this way as it is easier to debug the differences if the tests fail.

test/testMdastToVDOM.js
```diff
@@ -33,9 +33,7 @@ const assertTransformerYieldsDocument = (transformer, expected) => {
   // check equality of the dom, but throw assertion based on strings to visualize difference.
   const act = transformer.getDocument();
   const dom = new JSDOM(expected);
-  if (!eq(act, dom.window.document)) {
-    assert.equal(act.serialize(), dom.serialize());
-  }
+  assertEquivalentNode(act, dom.window.document);
```
tripodsan: same here. please keep the the logic

test/testHTMLFromMarkdown.js
koraa: Hi Tobias,

thanks for the review! The change was necessary to make the tests pass again;
assertEquivalentNode has machinery built in to display the contents of the dom elements,
so extra logic in here is not required :)

assertEquals should be able to use the same machinery IMHO, I am currently designing a method to let it display DOM elements in a generic way.

koraa:
@tripodsan Can we merge this?

test/testHTMLFromMarkdown.js
tripodsan: > built in to display the contents of the dom elements,

yes, but it's not nicely formatted. so the diff is very difficult to spot:

before:
image

after:
image


anyways...that's a minor detail for now.

PR #440, opened 2019-08-13
 feat(pipe): uniform steps (#377)

### LINKED ISSUE #377: RFC: uniform steps
## Overview
the current pipeline has a clear separation of `before`, `once` and `after` pipeline functions.
This allows to easily replace the `once` action, but requires the extension mechanism to inject other steps. this proposal tries to remove this distinction and make it easier to manipulate an existing pipeline. 

### 1. Uniform steps

Deprecate (or remove): `before()`, `once()`, `after()` in favour of a single `use()` method that constructs the pipeline. Since the old methods were used in sequence anyways, they can easily just be mapped to the`use()` method.

```ts
/**
 * Adds a processing function to this pipeline.
 * @param {pipelineFunction} f function to add.
 * @returns {Pipeline} this
 */
use(f:pipelineFunction):Pipeline;
```

Example:

```js
pipe
  .every(dump.record)
  .before(fetch).expose('fetch').when(hascontent)
  .before(parse).expose('parse')
  .before(html).expose('html')
  .once(cont)
  .after(type('text/html'))
  .after(addHeaders)
  .after(tohtml)
  .after(debug)
  .after(timer.report)
  .error(dump.report)
  .error(selectStatus)
```

just becomes:

```js
pipe
  .every(dump.record)
  .use(fetch).expose('fetch').when(hascontent)
  .use(parse).expose('parse')
  .use(html).expose('html')
  .use(cont)
  .use(type('text/html'))
  .use(addHeaders)
  .use(tohtml)
  .use(debug)
  .use(timer.report)
  .error(dump.report)
  .error(selectStatus)
```

### 2. Support replacing steps

Sometimes it might be convenient to modify an existing default pipeline. This is already marginally possible using `attach.before()` and `attach.after()`. However, it should also be possible to replace existing steps. 

```ts
/**
 * Replaces an existing handler of this pipeline. The function `fn` will be run in
 * the pipeline instead the function called `name`. If `fn` is {@code null}, the 
 * respective function is removed. 
 * If `name` does not exist, this method has no effect.
 * @param {string} name - name of the extension point (typically the function name).
 * @param {pipelineFunction} fn - a new pipeline step that will replace `name`.
 * @returns {Pipeline} this.
 */
replace(name:string, fn:pipelineFunction):Pipeline;
```

Example:

```js
module.exports.defaultPipe = pipe
  .use(fetch).expose('fetch').when(hascontent)
  .use(parse).expose('parse')
  .use(html).expose('html')
  .use(defaultRenderer).expose('render')
  .use(tohtml)
  .use(debug)
  .error(selectStatus)
```
some where else...
```js
const p = require('./defaultPipe');
p.replace('render', myCoolRenderer).run(...);
```

rofe:
Fixes #377 

Note that I slightly deviated from the proposed implementation in #377 in two places:
- I kept the `once()` method for now, mainly because of the special handling of the customer-specified continuation function as potential bearer of custom extensions. It felt cumbersome to check for extensions on every function that gets passed to `use()`. Maybe there's a simple and elegant way of doing that, but I didn't find one yet.
- I didn't implement a `pipeline.replace()` function, because it would violate the "seal" concept applied in `this.attach()`. Instead I went for a `this.attach.replace()` sub function and a check for a `replace` export on the continuation function.


tripodsan:
> * I kept the `once()` method for now, mainly because of the special handling of the customer-specified continuation function as potential bearer of custom extensions.

I don't understand. that's exactly the goal of #377 to get rid of special meaning of certain steps.

tripodsan:
> * I didn't implement a `pipeline.replace()` function, because it would violate the "seal" concept applied in `this.attach()`.

I think we can remove any sealing. it doesn't give any advantage.

src/pipeline.js
```diff
@@ -378,7 +387,7 @@ class Pipeline {
     };
 
     try {
-      await execFns([...this._pres, this._oncef, ...this._posts]);
+      await execFns([...this._steps]);
```
tripodsan:
```suggestion
      await execFns(this._steps);
```

src/pipeline.js
```diff
@@ -109,34 +105,33 @@ class Pipeline {
...
-      }
-      if (foundposts !== -1) {
-        this._posts.splice(foundposts + offset, 0, f);
-      }
-      if (foundpres === -1 && foundposts === -1) {
+      if (foundstep !== -1) {
+        if (offset === -1) {
+          this._steps.splice(foundstep, 1, f);
+        } else {
+          this._steps.splice(foundstep + offset, 0, f);
```
tripodsan: I would explicitely check for `<0` `==0` `>0`, otherwise someone can pass in `2` and messup...

rofe:
> so there is no special handling of the `cont` function anymore.

Got it. Should we also change the way develoeprs can register functions to extension points, then? Right now we look at an `ext` property on `this._oncef` where extensions are stored.

src/pipeline.js
rofe: Good point. I think I'd treat anything greater than 0 as 1, i.e. `after`

rofe:
> this will completely break backward compatibility

Ok, so no need to keep deprecated `before` and `after` around, either, I guess.

rofe:
> you don't need to actually replace the function, just invoke the replacement instead of the existing one. looking at the code, this is what you do ;-)

I'm not sure I understand... yes, we could keep a separate list of "overrides" and execute those instead of the original functions. But wouldn't that defeat the purpose of sealing the pipeline? Shouldn't we treat custom replacements the same way as before/after hooks?

src/pipeline.js
rofe: Actually, `attachGeneric` is a private method not accessible from the outside... But I guess better safe than sorry ;)

tripodsan:
> Shouldn't we treat custom replacements the same way as before/after hooks?

yes. your comment in the description was not clear enough - but then I looked at the code, and I think it is correct.

tripodsan:
> > so there is no special handling of the `cont` function anymore.
> 
> Got it. Should we also change the way developers can register functions to extension points, then? Right now we look at the properties `before` and `after` of the `this._oncef` where extensions are stored.

hmm, true.. I guess any `use()` function can add extension points ;-)

the problem is, that we want to be able to extend the pipe in the `pre.js` )(which is currently turned into the once function). if we don't have a dedicated `once` function anymore, we can either:

1. allow any `use()` function to add and extension point (which would be kind of confusing)
2. move the extend logic to the helix-cli, where we instrument the pipe.

src/pipeline.js
```diff
@@ -109,34 +105,38 @@ class Pipeline {
...
       // new function into the list, with the correct offset
-      if (foundpres !== -1) {
-        this._pres.splice(foundpres + offset, 0, f);
-      }
-      if (foundposts !== -1) {
-        this._posts.splice(foundposts + offset, 0, f);
-      }
-      if (foundpres === -1 && foundposts === -1) {
+      if (foundstep !== -1) {
+        if (offset < 0) {
```
tripodsan: this is counter intuitive: `<0` should be before, `==0` replace, and `>0` after.

rofe:
> > > so there is no special handling of the `cont` function anymore.
> > 
> > 
> > Got it. Should we also change the way developers can register functions to extension points, then? Right now we look at the properties `before` and `after` of the `this._oncef` where extensions are stored.
> 
> hmm, true.. I guess any `use()` function can add extension points ;-)
> 
> the problem is, that we want to be able to extend the pipe in the `pre.js` )(which is currently turned into the once function). if we don't have a dedicated `once` function anymore, we can either:
> 
> 1. allow any `use()` function to add and extension point (which would be kind of confusing)
> 2. move the extend logic to the helix-cli, where we instrument the pipe.

I would stay away from (1) and postpone (2). For now, I think I'd keep the `once()` function, but maybe rename it to `cont()` or `ext()` to better illustrate its purpose.

@trieloff what's your take here?

trieloff:
> 1. allow any `use()` function to add and extension point (which would be kind of confusing)

cool, but confusing, indeed.

> 2. move the extend logic to the helix-cli, where we instrument the pipe.

I don't like that.


My suggestion would be 1*:

*: we allow the use of arbitrary extensions in `use()`, but discourage it, e.g. by discovering when `before` or `after` are used in the second or later invocation of `use()` and either warning or throwing an error.

tripodsan:
> My suggestion would be 1*:

IMO there are 2 different usage patters or API's here. one is the pipeline API itself. it currently offers a `Pipeline.attach.before()` and `Pipeline.attach.after()` methods. they have no effect, once the pipeline steps are sealed. I think the signature is a bit weird, but I don't have a better option right now.

the other usage is what we offer in the defaults, eg the 

```
const htmlpipe = (cont, context, action) => {
...
  pipe
...
    .before(sanitize).when(paranoid)
    .once(cont)
    .after(type('text/html'))
...
  action.logger.log('debug', 'Running HTML pipeline');
  return pipe.run(context);
};
```

the `cont` here, is a more or less arbitrary decision to have a designated function for the _render_ step. also, the function also runs the pipe with the given context.

instead I propose that the default html pipe only returns the pipeline itself, but the user can then extend and run the pipe.

eg:

```
cont pipe = require('defaults/html.pipe.js').create();
pipe.attach.before('render', preFn);
pipe.attach.replace('render', htlRenderer);
pipe.run(context);
```

in order to simplify the attaching, we can have a 

```
pipe.attach(/* options */); 
```

which takes an object with `before`, `replace`, `after` configs, like in the current pre.js. so we can still attach the extension defined in the pre.js:

```
cont pipe = require('defaults/html.pipe.js').create();
pipe.attach(preFn);
pipe.attach.before('render', preFn);
pipe.attach.replace('render', htlRenderer);
pipe.run(context);
```

rofe:
From the implementation, it doesn't look like the `pipe.attach.*()` methods were ever meant to be public API. They are rather an implementation detail for one-off processing of the `before` and `after` extension objects on the `cont` function. This is also confirmed by the fact that the only [officially documented](https://github.com/adobe/helix-pipeline#using-extension-points) way to extend the pipeline is to export `before` and `after` objects in `html.pre.js`. 

The second part of this issue was about introducing a way to also replace pipeline steps. Sure, we could now also introduce an officially public `pipe.attach(obj)` method. But do we have known use cases so far where customers create their own pipelines and interact with the `pipe` object?

The more I think about it, the more I believe that the proposed changes to the extension mechanism would best be moved into a separate issue, and this issue should stay focused on uniforming pipeline steps: `pipe.use()` instead of `pipe.before(), `pipe.after()` and `pipe.once()`. Wdyt?

tripodsan:
> The more I think about it, the more I believe that the proposed changes to the extension mechanism would best be moved into a separate issue,

well, just renaming before/after is not the main focus of the API change. rather that the `cont` / `once()` function should not have a special meaning anymore. 

so the changes would be:

**pipeline**
- add `use()` method
- deprecate `before()`, `after()` and `once()` and just invoke `use()`
- add `attach.replace()`
- create new `attach(options)` method that takes a `before/after/replace` object.
- change all default pipelines:
   - remove all `cont` functions
   - do not run pipeline, just create it

**helix-cli**
- handle attach logic (i.e. call `pipe.attach(preFn)` )
- run pipeline explicitly

rofe:
ok, i basically have the changes in the pipeline ready:
- once(), before() and after() replaced by use()
- first function with extension points wins, use() throws error on next
- attach() no longer seals the pipeline so it can be called from outside

will create a related issue for the `helix-cli` changes

src/pipeline.js
```diff
@@ -90,53 +88,51 @@ class Pipeline {
 
     // function chain that was defined last. used for `when` and `unless`
     this._last = null;
```
tripodsan: I think the `this._last` is not used anymore

src/pipeline.js
```diff
@@ -90,53 +88,51 @@ class Pipeline {
...
+    // function that is executed once and can register extensions
     this._oncef = null;
-    // functions that are executed after
-    this._posts = [];
+    // step functions to execute
+    this._steps = [];
     // functions that are executed before each step
     this._taps = [];
 
     this.attach = (ext) => {
```
tripodsan: maybe _promote_ this to a proper method and add documentation?

src/pipeline.js
```diff
@@ -147,38 +143,41 @@ class Pipeline {
...
-    this._posts.push(f);
-    this._last = this._posts;
+    this._steps.push(f);
+    this._last = this._steps;
+    // check for extensions
+    if (f && (f.before || f.replace || f.after)) {
+      if (typeof this._oncef === 'function') {
+        throw new Error(`Step '${this._oncef.alias}' already registered extensions for this pipeline, refusing to add more with '${f.alias}'.`);
+      }
+      this._oncef = f;
```
tripodsan: would rename this to `sealed`....

tripodsan:
well, ok...let's do this in 2 steps.

this PR will just add uniform steps, but leaves the defaults untouched...

then we create an additional issue that removes the `run()` from the default pipelines.

src/pipeline.js
rofe: Yes it is used in `when()`

src/pipeline.js
rofe: I don't understand: the method behind it is the new `use()`. I thought about renaming the property to `_use`, but found it a bit confusing.

src/pipeline.js
rofe: You mean the `_oncef`?

src/pipeline.js
rofe: Ah, you mean I could use `this._steps` directly there, now that we only have one sequence of functions... good point!

rofe: No, we still have 2 queues: `_steps` and `_taps` (used by `pipe.every()`)

src/pipeline.js
rofe: Oops, never mind: you meant `attach()` - of course!

src/pipeline.js
rofe: or `_cont`? In essence, this is what's left of the continuation function.

rofe:
If we remove all `cont` functions from the default pipelines, how would helix-cli know where to insert the `preFn`?

tripodsan:
> If we remove all `cont` functions from the default pipelines, how would helix-cli know where to insert the `preFn`?

because we have a default (nop) step called `render` and can replace it.

src/pipeline.js
tripodsan: hmm...I see. maybe rename it to `_overrides` or `_plugins` or `_attachments`

rofe:
I see. Let's create a separate issue for this change.

rofe:
https://github.com/adobe/helix-cli/issues/1125 is implemented in helix-cli, adding support replacing pipeline steps. With that this PR addresses all original points of #377 (uniform steps with `use()` and add `replace` support).

@trieloff @tripodsan any additional concerns with merging?

src/pipeline.js
rofe: but in most cases, it's the `pre` (aka continuation) function...

rofe: I was wrong, thanks @tripodsan. Renamed it to `_attachments` now.

PR #454, opened 2019-09-11
feat(pipe): added queryString property

MarquiseRosier:
https://github.com/adobe/helix-pipeline/issues/366


src/utils/openwhisk.js
```diff
@@ -66,6 +66,7 @@ function extractClientRequest(action) {
     selector: request.params.selector || '',
     url: `${reqPath}${query}`,
     rootPath: request.params.rootPath,
+    queryString: request.params.params,
```
tripodsan: according to the schema example and the [spec](https://url.spec.whatwg.org/#dom-url-search), the `queryString` property should either be empty or start with a `?`.

PR #455, opened 2019-09-12
Relax Frontmatter Parsing

### LINKED ISSUE #441: markdown parser should be more resilient
problems during markdown parsing should not throw a error and abort the pipeline.

eg:

```
Error: Found ambigous frontmatter fence: No empty line before the block! Make sure your frontmatter blocks contain no empty lines and your horizontal rules have an empty line before AND after them.
```

trieloff:
- no longer throw exceptions (we need to think about how to report this)
- do not require non-empty lines in the front matter block at the start of the document (same rules for other blocks)

fixes #441

### DIFF
--- a/src/html/parse-frontmatter.js
+++ b/src/html/parse-frontmatter.js
@@ -18,6 +18,7 @@ const {
 } = require('ferrum');
 
 const { assign } = Object;
+const production = require('../utils/is-production');
 
 /**
  * Given an mdast tree and it's text representation, this finds all
@@ -150,7 +151,6 @@ const findFrontmatter = (mdast, str) => {
   const toIgnore = (n) => !n || decentHead(n) || decentHr(n);
 
   const procwarnigns = map(([fst, last]) => {
-    console.log('procwarnings', fst, last);
     const src = str.slice(fst.offStart, last === null ? undefined : last.offEnd);
 
     const warn = (cause, prosa) => ({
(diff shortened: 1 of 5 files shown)


src/html/parse-frontmatter.js
```diff
@@ -301,7 +302,11 @@ const parseFrontmatter = ({ content = {} }) => {
         join('\n'),
       );
 
-      throw new FrontmatterParsingError(`${warning}\n${sourceref}`);
+      const err = new FrontmatterParsingError(`${warning}\n${sourceref}`);
```
koraa: Why not just always warn?

src/html/parse-frontmatter.js
```diff
@@ -173,7 +174,7 @@ const findFrontmatter = (mdast, str) => {
         'Found ambigous frontmatter fence: No empty line after the block! '
           + 'Make sure your frontmatter blocks contain no empty lines '
           + 'and your horizontal rules have an empty line before AND after them.');
-    } else if (src.match(re(`\\n${hspace}*\\n`))) {
+    } else if (src.match(re(`\\n${hspace}*\\n`)) && fst.idx > 0) {
```
koraa: If you change this, you should also change the documentation appropriately.

src/html/parse-frontmatter.js
trieloff: Because you don't want to fill up the error logs in production with things that are not relevant at that time.

koraa:
You could leave me a day to review…
Also did you address my feedback about using an option about relaxing rule 2?

trieloff:
I'm trying to avoid options until they have proven a practical necessity.

PR #460, opened 2019-09-19
docs(frontmatter): update API docs to describe parsing rules

trieloff:


src/html/parse-frontmatter.js
```diff
@@ -46,6 +46,14 @@ const production = require('../utils/is-production');
  *   arrays are not considered to be frontmatter in this context
  * - The thematic break must be made up of three dashes
  *
+ * For frontmatter blocks at the beginning of the document, the rules are a bit
+ * looser, as there is less risk of confusion:
+ * - There must be the start or end of the document or an empty line (`\n\n`)
```
koraa: This is now redundand (if the FM block is at the start of the document, of course it's at the startof the document)

tripodsan: I agree with @koraa...this paragraph is almost identical to the one above....

PR #466, opened 2019-09-30
Frontmatter error

MarquiseRosier:


src/html/parse-frontmatter.js
```diff
@@ -189,7 +189,7 @@ const findFrontmatter = (mdast, str) => {
           + 'and your horizontal rules have an empty line before AND after them.');
     }
 
-    const txt = str.slice(fst.offEnd, last.offStart);
+    const txt = str.slice(fst.offEnd, last === null ? undefined : last.offStart);
```
trieloff:
```suggestion
    const txt = str.slice(fst.offEnd, !last ? undefined : last.offStart);
```

that may be a bit shorter. I'm not sure if `null === undefined`.

MarquiseRosier:
I'm not going to merge this yet. I found an error in my local tests; and I want to make sure it's not in this pr

PR #473, opened 2019-10-03
fix(parse): fix empty extension case and add tests

kptdobe:
PR for https://github.com/adobe/helix-pipeline/issues/472

Did not find how to do that simply (and user readable) with Ferrum. Back to good old `if`.

Also added tests for uncovered area.

### DIFF
--- a/src/html/parse-markdown.js
+++ b/src/html/parse-markdown.js
@@ -19,7 +19,8 @@ function parseMarkdown(context, action) {
   const body = setdefault(content, 'body', '');
 
   const request = setdefault(context, 'request', {});
-  const extension = setdefault(request, 'extension', 'html');
+  if (!request.extension) request.extension = 'html';
+  const { extension } = request;
 
   action.logger.debug(`Parsing markdown from request body starting with ${body.split('\n')[0]}`);
   content.mdast = unified().use(remark).parse(body);
--- a/test/testParseMarkdown.js
+++ b/test/testParseMarkdown.js
@@ -10,6 +10,7 @@
  * governing permissions and limitations under the License.
  */
 /* eslint-env mocha */
+const assert = require('assert');
 const { Logger } = require('@adobe/helix-shared');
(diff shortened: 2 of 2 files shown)


src/html/parse-markdown.js
```diff
@@ -19,7 +19,8 @@ function parseMarkdown(context, action) {
   const body = setdefault(content, 'body', '');
 
   const request = setdefault(context, 'request', {});
```
trieloff: I'm not sure I like either variant. Both have side effects (modify context.request) that are hidden in the parse-markdown step. Would there be any harm in simply extracting the extension as a local const?

kptdobe: `modify context.request` was already there before. I did not change that part.
Pb is that there is no extension set, thus it has to be set somewhere anyway. And before previous refactoring, it was set in the `make-html` (not better, I agree).

I started to refactor the whole thing (extract those default setters and put them before). And then reverted the changes because all the tests were failing and this was a huge change that would have a big impact. Out of scope for this PR. I would be a topic for another issue. I would fix the current real issue we have here and create a follow up issue for the refactoring.

trieloff: Just a comment in general, your PR is fine to be merged in my eyes.

trieloff: Also, we coerce defaults for secrets using the schema and could do the same for other context values.

tripodsan: I think this is completely at the wrong place. IMO action and context objects should be setup before.

tripodsan:
@kptdobe @trieloff I think this is wrong and I would revert it, and rather fix the code that expects the extension to be not empty.

kptdobe:
As mentioned above, I did not change the existing logic, I just fixed a pure coding issue. As you have seen and since changing this logic is out of scope for the current issue, I created a follow up issue and I have taken the discussion in https://github.com/adobe/helix-pipeline/issues/474.

PR #490, opened 2019-10-16
Added table of contents for easier navigation

MohanSha:
Added table of contents for easier navigation


tripodsan:
thanks. but this looks like an automated task, so marking it invalid.

trieloff:
@MohanSha if you'd like to contribute a script that automatically updates the TOC as part of the documentation build, you are more than welcome. We just don't think that a one-time injection of a TOC will be sustainable after October.

Issue #499, opened 2019-10-21
links with formatting produce 'undefined'


tripodsan:
eg:

```
[<u>**Adobe Illustrator: New Features, Pro Tips and Time-savers – Paul Trani, Principal Worldwide Creative Cloud Evangelist, Adobe**</u>](https://events.rainfocus.com/widget/adobe/am19/sessioncatalogmax?search=s6309)
```

Github produces:

[<u>**Adobe Illustrator: New Features, Pro Tips and Time-savers – Paul Trani, Principal Worldwide Creative Cloud Evangelist, Adobe**</u>](https://events.rainfocus.com/widget/adobe/am19/sessioncatalogmax?search=s6309)

is rendered as:

```
<a href="https://events.rainfocus.com/widget/adobe/am19/sessioncatalogmax?search=s6308"><u>undefined</u></a>
```

tripodsan:
@ramboz  can you have a look? it's a problem of `VDOMTransformer.sanitizeInlineHTML()`.
I added 2 test cases to #501

tripodsan:
closing as won't fix for now.

PR #501, opened 2019-10-22
fix(link): tenative fix for raw elements

tripodsan:
see #499 

This is a partial fix for cases where a _raw_  elements have markdown nodes as children.


src/utils/mdast-to-vdom.js
```diff
@@ -223,7 +224,7 @@ class VDOMTransformer {
             const last = stack.pop();
             let html = '';
             for (let j = last; j <= i; j += 1) {
-              html += node.children[j].html || node.children[j].value;
+              html += node.children[j].html || node.children[j].value || toString(node.children[j]);
```
ramboz: From my understanding, this will strip any html in the children.
What about something like:
```js
if (node.children[j].html || node.children[j].value) {
  html += node.children[j].html || node.children[j].value;
}
else {
  const dom = new JSDOM();
  const doc = dom.window.document;
  const frag = toDOM(doc, node.children[j], { fragment: true });
  html += frag.outerHTML;
}
```

The `sanitize` step in the pipeline would still apply after this if I'm not mistaken, so that should also address possible XSS.

tripodsan: this, this is a tentative fix... I think the sanitize dom just needs to be smarter. I would like to avoid any reparsing of the html.

test/testHTMLFromMarkdown.js
```diff
@@ -482,4 +482,24 @@ describe('Testing Markdown conversion', () => {
       SANITIZE_DOM: true,
     });
   });
+
+  it('formats links correctly', async () => {
+    process.env.__OW_ACTIVATION_ID = '1234';
+    await assertMd(
+      '[<u>**Adobe**</u>](https://www.adobe.com)',
+      '<body><p><a href="https://www.adobe.com"><u><strong>Adobe</strong></u></a></p></body>', {
```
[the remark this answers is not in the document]
trieloff: Ideally, you would recurse into the children of the link, but stripping markup is still better than `undefined`.

tripodsan:
closing as won't fix for now.

PR #505, opened 2019-10-22
fix(html): consistent hlx props removal

tripodsan:


src/defaults/html.pipe.js
```diff
@@ -77,7 +77,7 @@ const htmlpipe = (cont, context, action) => {
     .use(cache).when(uncached)
     .use(key)
     .use(tovdom).expose('post') // start HTML post-processing
-    .use(removeHlxProps).when(() => production())
+    .use(removeHlxProps).expose('removeHlxProps')
```
trieloff: I'd prefer a lowercase extension point name for consistency sake.

```suggestion
    .use(removeHlxProps).expose('cleanup')
```

ramboz: Is there a "debug" mode in the pipeline? If yes, would it make sense to leave them in that case?

Issue #522, opened 2019-11-08
Accessibility


rofe:
For screen readers to be able to efficiently navigate the HTML output of the pipeline - specifically `VDOMTransformer.getDocument()` - it should be accessible with landmarks, title and ARIA attributes by default. (For landmarks, we might want to consider using HTML5 tags like `main`, `nav`, etc. which would eliminate the need to introduce `role` attributes.)

References:
- https://www.w3.org/WAI/fundamentals/accessibility-principles/
- https://a11y-101.com/

trieloff:
I feel like `div` vs. `main` or `nav` or `section` is reopening an old discussion, but I agree that we should have the correct accessibility markers by default.

rofe:
Do you have a reference to the "old discussion"? I may not be the only one wondering why we aren't leveraging HTML5 out of the box.

trieloff:
#279 is probably the best pointer.

ramboz:
For what it's worth, my opinion on this is that Helix Pages should probably offer WCAG 2.0 AA level out of the box, same for the default HTL in the Helix demo site.

On the other hand, I'd expect devs using Helix directly (i.e. HTL & pre.js) to take care of all accessibility/SEO/microdata on their own mostly. Helix should not impose any semantics on the devs and let them have full control of the markup.

tripodsan:
@rofe is there still something we should improve / change on the default HTML output of the pipeline ?

rofe:
After the latest accessibility review of a helix proejct, I think we're good for now.

PR #528, opened 2019-11-11
fix(icons): use title from SVG file

### LINKED ISSUE #527: [icons] use <title> tag from SVG file
Currently, we write:
```html
<svg class="icon icon-twitter">
    <use xlink:href="/icons/bla.svg">
        <title>bla</title>
    </use>
</svg>
```
Instead, it should be:
```html
```html
<svg class="icon icon-twitter">
    <use xlink:href="/icons/bla.svg"></use>
</svg>
```
since the `<title>` tag should come from the SVG symbol in the referenced file.

See also https://css-tricks.com/svg-symbol-good-choice-icons/#article-header-id-1

rofe:
Fix #527


rofe:
I can see the `<title>` tag in the shadow DOM when inspecting the element, but the browser doesn't display any tool tip unfortunately...

rofe:
@davidnuescheler @trieloff wdyt?

trieloff:
https://css-tricks.com/accessible-svgs/ (scroll to "Inline SVG") has some suggestions for working around the browser bugs.

rofe:
Thanks for the pointer. However I don't think we want to inline SVG, do we? We could, using the same ESI trick we use for other statics... 

I explicitly decided against adding `aria-*` and `role` attributes because I think we should tackle this holistically in #522.

Issue #523, opened 2019-11-08
SVG icon sprites


rofe:
As originally mentioned by @davidnuescheler in #500, this enhancement request is for adding support for SVG sprites, so all icons of a site could be stored in a single SVG file and referenced by ID, e.g.:

Sprite URL: `/icons.svg` (configurable or by convention)
```html
<svg class="icon icon-twitter">
    <use xlink:href="/icons.svg#twitter">
        <title>twitter</title>
    </use>
</svg>
```

rofe:
@davidnuescheler should we move this to `helix-pipeline-service` or close won't fix?

PR #529, opened 2019-11-12
Icon config

rofe:
Proposal to enable #523 without breaking backward compatibility.


src/schemas/secrets.schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "meta:license": [
-    "Copyright 2018 Adobe. All rights reserved.",
+    "Copyright 2019 Adobe. All rights reserved.",
```
trieloff: I don't think you have to update the year here. The initial claim has been made in 2018, what has been added in 2019 is only incremental.

rofe: OK

rofe:
> The code looks good, but do we really _need_ the config?

If we want to enable #523 without disabling single SVG icon files at the same time, I thought a config would be the least intrusive/breaking way to have both. But I'm open for alternative solutions.

trieloff:
Ok

rofe:
I agree that we'll need a solution that works for helix-pages. Closing this PR.

PR #530, opened 2019-11-12
docs(contributing): what is a breaking change

trieloff:
see https://github.com/adobe/helix-pipeline/issues/509#issuecomment-552963032


CONTRIBUTING.md
```diff
@@ -55,6 +55,15 @@ $ git add package.json
...
 
+### Breaking Changes
+
+`npm run commit` will ask you if a change is a breaking change. This is a non-exhaustive list of what we consider breaking changes for `helix-pipeline`
+
+* any change that will remove or rename HTML attributes in the generated HTML output
+* any change that will remove or rename HTML tags in the generated HTML output
+* any change that will remove or rename configuration settings in `secrets`
+* any change that will remove or rename properties exported through `index.js module.exports`
+
```
rofe: We might also add:
* any change that will alter the structure of the generated HTML output

trieloff: I think the first two lines cover that more succinctly. It's not that any change is a breaking change, only changes that remove or rename tags or attributes. Your selectors should still work for additive changes to the HTML.

rofe: I was thinking about something like
```html
<div class="main">
  <p>Foo</p>
</div>
```
with the following customer CSS:
```css
div.main > p {
  font-size: 1.5em;
}
```
We might change the structure of the HTML by adding tags:
```html
<div class="main">
  <div>
    <p>Foo</p>
  </div>
</div>
```
This will break the DOM contract.

CONTRIBUTING.md
```diff
@@ -55,6 +55,15 @@ $ git add package.json
...
 
+### Breaking Changes
+
+`npm run commit` will ask you if a change is a breaking change. This is a non-exhaustive list of what we consider breaking changes for `helix-pipeline`
+
+* any change that will remove or rename HTML attributes in the generated HTML output
+* any change that will add, remove, or rename HTML tags in the generated HTML output
+* any change that will remove or rename configuration settings in `secrets`
+* any change that will remove or rename properties exported through `index.js module.exports`
+
```
tripodsan:
```suggestion
* any change that alters the schemas in a non-backward compatible way
```

trieloff:
```suggestion
* any change that adds required properties to the schemas
* any change that removes or renames properties in the schemas
```

PR #532, opened 2019-11-14
feat(log): use helix-log instead of winston

tripodsan:
- uses openwhisk-action-utils to setup logging (gives coralogix support)
- uses helix-log under the hood
- winston log interface compatible
- removed default file logger of 'logs/pipeline.log'.


tripodsan:
there is still one test failing using `nock` when executing all tests at once. executing the test itself works.
I'll need some more time to fix that, but if you could already review the changes, that would be great.

kptdobe:
Smoke tests also fail: https://circleci.com/workflow-run/2535ca55-f1a5-4a0b-8f04-eeef0de0a28f
Plenty of `[hlx] error: [object Object]` while running a project.

kptdobe:
Ah I think the pb of the smoke tests is coming from the simulator which also need to be migrated.

kptdobe:
Please test the changes in "production" because I tried and I still do not get Coralogix logs. But this might be due to my config / setup... like Coralogix key missing in my namespace... I am investigating but for now, cannot find an explanation.

kptdobe:
Just tested and no pipeline logs is sent to Coralogix.

tripodsan:
> Just tested and no pipeline logs is sent to Coralogix.

and if you test locally? i.e. executing the `html.bundle.js` ?

kptdobe:
humm... I can still find some `winston` references in that `html.bundle.js`. Need to investigate further.

kptdobe:
ok, found why I did not have the latest version (my project had set a dependency to older helix-pipeline). But with latest code, still no logs in Coralogix.

tripodsan:
> humm... I can still find some `winston` references in that `html.bundle.js`. Need to investigate further.

yes. they come from helix-shared...

tripodsan:
> ok, found why I did not have the latest version (my project had set a dependency to older helix-pipeline). But with latest code, still no logs in Coralogix.

what happens if you execute the bundle locally. with the node REPL ?

kptdobe:
How do I do this ?

tripodsan:
```
$ hlx package
$ node -p "require('./.hlx/build/src/html.bundle.js').main({CORALOGIX_API_KEY:'sdf'}).then(console.log).catch(console.error)"
```

kptdobe:
Works!

For the record: 

My deployed action was not logging anything to Coralogix because the `CORALOGIX_API_KEY` was missing during the deployment:
```
hlx deploy --default CORALOGIX_API_KEY XXXX
```
fixed the issue.

Thanks.

PR #561, opened 2020-01-14
fix(md): ignore non-matching lineline elements

### LINKED ISSUE #557: Unmatched closing html tag in content triggers a 500
In a md file, add an html closing tag like `</p>`
-> render the page: error 500.

If you add an opening html tag like `<p>`, it works fine.

I think we should just get rid of the invalid html instead of throwing an error.

tripodsan:
fixes #557


src/utils/mdast-to-vdom.js
```diff
@@ -220,7 +220,7 @@ class VDOMTransformer {
       if (child.type === 'raw') {
         if (child.value.firstElementChild === null) {
           if (stack.length === 0) {
-            throw new Error(`no matching inline element found for ${child.html}`);
+            // ignore unmatched inline elements
```
ramboz: Would it make sense to somehow log a info/warning to track these, so content authors can ultimately fix the issue?

tripodsan: yes, however, I don't think that authors will ever see those messages.

PR #564, opened 2020-01-15
feat(html): rewrite image URL for Azure Blob store

trieloff:
When using the Azure Blob store, for instance through `helix-googledocs2md` or `helix-word2md`, image links point directly to the Azure Blob store. This change replaces these URLs in the HTML output with links to the `hlx_*.jpg`, i.e. Blob Proxy support, so that images can be resized and delivered from the same domain as the main content.

fixes https://github.com/adobe/helix-word2md/issues/128 fixes https://github.com/adobe/helix-googledocs2md/issues/133 closes https://github.com/adobe/helix-word2md/pull/129 requires https://github.com/adobe/helix-publish/pull/278 supports https://github.com/davidnuescheler/theblog/issues/43

### DIFF
--- a/src/utils/image-handler.js
+++ b/src/utils/image-handler.js
@@ -18,7 +18,10 @@ function image() {
       const patchednode = {
         ...node,
       };
-      patchednode.url = `${node.url.replace(/^https:\/\/hlx\.blob\.core\.windows\.net\/external\//, '/hlx_')}.jpg`;
+      const [filename, fragment] = node.url.split('/').pop().split('#');
+      const fragmentextension = fragment ? fragment.split('.').pop() : undefined;
+      const extension = fragmentextension || 'jpg';
+      patchednode.url = `/hlx_${filename}.${extension}`;
       return fallback(h, patchednode);
     }
     return fallback(h, node);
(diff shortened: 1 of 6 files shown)


src/utils/image-handler.js
```diff
@@ -0,0 +1,28 @@
...
+const fallback = require('mdast-util-to-hast/lib/handlers').image;
+
+function image() {
+  return function handler(h, node) {
+    // rewrite blob links to use Fastly
+    if (/^https:\/\/hlx\.blob\.core\.windows\.net\/external\//.test(node.url)) {
+      const patchednode = {
+        ...node,
+      };
+      patchednode.url = `${node.url.replace(/^https:\/\/hlx\.blob\.core\.windows\.net\/external\//, '/hlx_')}.jpg`;
```
tripodsan: I find it sub-optimal, that we lose the extension. we could convey it by using a `#hash`, eg:

```
https://hlx.blob.core.windows.net/external/b7a97d3fc2c553bacf17a6e0d69d3e6e7fbab63b#image.png
```

@trieloff WDYT?

trieloff: Yes, that would be easy to do.

PR #569, opened 2020-01-20
chore(dep): move renovatebot to monday afternoon

rofe:


tripodsan:
btw: I think renovate also allows for shared config. https://docs.renovatebot.com/config-presets/
should we create a `helix-renovatetc` ?

rofe:
> or simply use a shared one, eg in helix-shared: github>adobe/helix-shared

I like it! This PR now depends on https://github.com/adobe/helix-shared/pull/230 and will make `helix-pipeline` the guinea pig for using a shared config...

PR #590, opened 2020-02-19
fix(embed): fix external embed service URL construction

### LINKED ISSUE #559: embed creates wrong link in html
for example:
```
https://adobeioruntime.net/api/v1/web/helix/helix-services/embed@v1https://embeds.project-helix.io/jekyll_embed?params=with%20params
```
(missing slash after `@v1`)

also  see https://github.com/adobe/helix-cli/pull/1255/files

trieloff:
the embed service expects the to-be-embedded URL to be passed after a trailing slash. The code did
not add the trailing slash, making it impossible for Adobe I/O Runtime to find the correct action to
execute. This change adds the correct slash again.

fixes https://github.com/adobe/helix-embed/issues/145 fixes #559

### DIFF
--- a/src/utils/embed-handler.js
+++ b/src/utils/embed-handler.js
@@ -20,7 +20,7 @@ function embed({ EMBED_SERVICE } = {}) {
     const { url } = node;
     const props = {
       // prepend the embed service for absolute URLs
-      src: (URI.parse(url).reference === 'absolute' ? EMBED_SERVICE : '') + url,
+      src: (URI.parse(url).reference === 'absolute' ? `${EMBED_SERVICE}/` : '') + url,
     };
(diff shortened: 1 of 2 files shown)


src/utils/embed-handler.js
```diff
@@ -20,7 +20,7 @@ function embed({ EMBED_SERVICE } = {}) {
     const { url } = node;
     const props = {
       // prepend the embed service for absolute URLs
-      src: (URI.parse(url).reference === 'absolute' ? EMBED_SERVICE : '') + url,
+      src: (URI.parse(url).reference === 'absolute' ? `${EMBED_SERVICE}/` : '') + url,
```
tripodsan: this could be more robust and deal with URLs with or without a trailing slash.....

Issue #516, opened 2019-10-30
Declaratively Configurable Markup


trieloff:
This is mainly related to the `helix-pages` use case, but should be implemented straight in the pipeline.

The HTML generated by the pipeline is intentionally minimalistic. As @anfibiacreativa pointed out during the hackathon, minimalistic markup can lead to maximalistic CSS selector expressions and it would be great to give developers a way to influence the generated markup.

With the `VDOMTransformer.match` API, this ability exists, but it comes with three drawbacks:
- lacking a `pre.js` in `helix-pages`, it cannot be used in the most common scenario
- being an imperative API, it cannot be safely run in a shared environment
- the API requires advanced understanding of the `unifiedjs`, `hyperscript` and DOM APIs, and requires recursion, which can pose a hurdle for most developers

In practice, most scenarios require one of the following:
1. **wrap** the generated HTML element in one or more developer-defined elements
2. **inject** `class` attribute values into the `class`attribute of the generated HTML element
3. **add** other attributes and values into the generated HTML element

These use cases can be solved using a similar approach we use for [query definitions](https://github.com/anfibiacreativa/helix-norddal/blob/master/helix-index.yaml).

Imagine a `markup.yaml` (https://github.com/adobe/helix-shared/issues/190) file like this:

```yaml
#
# markup configuration for theblog
#
version: 1
​
markup:
  images-in-gallery:
    match: section.is-gallery image
    wrap: <div class="gallery-image">
  external-links:
    match: link[href!="https://"]
    class: external
  last-section:
    match: section:last-of-type
    attribute:
      data-type: footer
```

The `match` expression would work on the MDAST tree, and the `wrap`, `class`, or `attribute` directives would apply to the generated DOM nodes.

davidnuescheler:
i think it would be great to support css selectors level 4

trieloff:
> i think it would be great to support css selectors level 4

But then they'd work on the DOM, not the MDAST, correct?

tripodsan:
> The `match` expression would work on the MDAST tree, and the `wrap`, `class`, or `attribute` directives would apply to the generated DOM nodes.

how would that work? this isn't very intuitive. Wouldn't it be better to match the DOM instead?

davidnuescheler:
[...]

PR #592, opened 2020-02-19
feat(fstab): add fstab support

tripodsan:
**Note**: This is a bit a bigger PR, due to the fact, that a lot of tests needed to be adjusted, since the new helix-fetch causes problems with pollysjs/nock.

- adding new `action.downloader` that can be used to schedule downloads for `fstab.yaml`, 
 `markup.yaml`, etc.  (/cc @ramboz )
- The `Dowloader` uses helix-fetch by default
- added new steps in html, xml, json pipeline: `fetch-fstab` and `fetch-external`


test/testDownloadManager.js
```diff
@@ -0,0 +1,291 @@
...
+      mgr.fetchGithub({
+        owner: 'adobe',
+        repo: 'test-repo',
+        ref: 'development',
+        path,
+        id: path,
+      });
+    });
+
+    // todo: this doesn't seem to work. maybe nock cannot intercept this. but why with github ?
```
tripodsan: @stefan-guggisberg Somehow, I can't get `nock` to intercept `www.example.com`. but it works for `raw. githubusercontent.com`. I don't understand the difference :-) I even tried to force `http1` protocol...but it just doesn't work :-)

src/utils/DownloadManager.js
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    if (!('timeout' in options)) {
+      options.timeout = this.timeout;
+    }
+
+    if (!('cache' in options)) {
+      options.cache = request.headers ? request.headers['Cache-Control'] : '';
+    }
+    options.cache = '';
```
tripodsan: @stefan-guggisberg when I comment this line, the `testHTML.js` tests don't terminate....probably related to helix-fetch.
also: is there a way to easily disable the cache? I think it confuses pollyjs.

tripodsan:
@stefan-guggisberg using helix-fetch together with pollyjs/nock is really a problem, as they can't intercept the requests. 

I think it would be cool, if helix-fetch had the option to use the 'normal' http1 adapter by default, at least for testing.

stefan-guggisberg:
> @stefan-guggisberg using helix-fetch together with pollyjs/nock is really a problem, as they can't intercept the requests.
> 
> I think it would be cool, if helix-fetch had the option to use the 'normal' http1 adapter by default, at least for testing.

@tripodsan I created an issue: https://github.com/adobe/helix-fetch/issues/8

src/defaults/html.pipe.js
```diff
@@ -59,6 +61,8 @@ const htmlpipe = (cont, context, action) => {
     .every(dump.record)
     .every(validate).when((ctx) => !production() && !ctx.error)
     .every(timer.update)
+    .use(fetchFstab).expose('fstab')
```
ramboz: From what I remember, @trieloff wanted to fetch in parallel instead of in sequence to reduce execution time.
So not sure if that has changed now, but I'll need to align #516 according to the consensus

ramboz:
@tripodsan 
> using helix-fetch together with pollyjs/nock is really a problem, as they can't intercept the requests.

I had issues as well but ultimately got it working during the hackathon. Anything in particular that fails on your side?

src/defaults/html.pipe.js
tripodsan: this step only schedules the donwload but doesn't wait for it....

tripodsan:
> I had issues as well but ultimately got it working during the hackathon. Anything in particular that fails on your side?

I don't think that pollyjs intercepts the requests. I turned my network off, and non of the tests worked anymore....

ramboz:
@tripodsan In my experience, polly was intercepting, but somehow fetch-h2 (used by helix-fetch) was still trying to resolve DNS and failed when offline. So yes, tests are not working offline, but polly can still mock them properly as far as I could tell. Just not the DNS resolution apparently.

@stefan-guggisberg Does that help narrow down the issue for adobe/helix-fetch#8?

tripodsan:
> @ramboz  In my experience, polly was intercepting, but somehow fetch-h2 (used by helix-fetch) was still trying to resolve DNS and failed when offline. So yes, tests are not working offline, but polly can still mock them properly as far as I could tell. Just not the DNS resolution apparently.

this is not what I observed. polly (which uses nock) cannot intercept requests using the http2 module.
btw, I didn't disable the entire network stack for my tests, just blocked the outgoing connections to raw.github etc...

ramboz:
@tripodsan hmm... Maybe I ended up in the http1 fallbacks and never actually had Polly working on the http2 calls. How do you enforce http2 in the tests?
Asking because I want to make sure my tests are consistent with the approach you followed here

tripodsan:
> Maybe I ended up in the http1 fallbacks and never actually had Polly working on the http2 calls. How do you enforce http2 in the tests?

IMO, there is not way (yet) to force http1. but adobe/helix-fetch#8 will address it

stefan-guggisberg:
> @tripodsan hmm... Maybe I ended up in the http1 fallbacks and never actually had Polly working on the http2 calls. How do you enforce http2 in the tests?
> Asking because I want to make sure my tests are consistent with the approach you followed here

What url's are you using? `https://raw.github.com` e.g. is *not* supporting HTTP/2, i.e. `helix-fetch` is not using the `http2` module. It uses `https` for this url.

`nock` (and `pollys`) doesn't (yet) intercept `http2` calls.

tripodsan:
> What url's are you using? `https://raw.github.com` e.g. is _not_ supporting HTTP/2, i.e. `helix-fetch` is not using the `http2` module. It uses `https` for this url.

how does it know that raw.github.com is not using http2? it has to make a request, and then it will fallback to http. but when I disable/block the network, the tests fail, since it cannot correct at all.

> `nock` (and `pollys`) doesn't (yet) intercept `http2` calls.

yes, this is the problem. I cannot mock reliably a backend.. only the ones that that don't support http2 and only if I don't block the calls. i.e. when raw.github.com suddenly starts to support http2, then all our tests will fail.

test/testDownloadManager.js
stefan-guggisberg: here's the difference ;)

https://http2.pro/check?url=https%3A//www.example.com/
https://http2.pro/check?url=https%3A//raw.githubusercontent.com/

src/utils/DownloadManager.js
stefan-guggisberg: `options.cache = 'no-store';` disables the cache.

stefan-guggisberg:
> how does it know that raw.github.com is not using http2? it has to make a request, and then it will fallback to http. but when I disable/block the network, the tests fail, since it cannot correct at all.

Initially It opens a socket to the server and uses ALPN to negotiate the supported protocols. This requires DNS resolution. `nock` intercepts requests at the `http` module level but it doesn't intercept raw socket I/O. This explains the test failures.

stefan-guggisberg:
> In my experience, polly was intercepting, but somehow fetch-h2 (used by helix-fetch) was still trying to resolve DNS and failed when offline. So yes, tests are not working offline, but polly can still mock them properly as far as I could tell. Just not the DNS resolution apparently.

`fetch-h2` initially opens a socket to the server and uses ALPN to negotiate the supported protocols. This requires DNS resolution. `nock` (and therefore `polly`) intercepts requests at the `http/https` module level but it doesn't intercept raw socket I/O.

src/html/fetch-external.js
```diff
@@ -0,0 +1,114 @@
...
+      uri: url.href,
+      // ump the timeout a bit, since the external script might take a while to render
+      options: {
+        timeout: 10000,
+      },
+      errorOn404: true,
+    });
+
+    // store extra source location if present
+    const sourceLocation = resp.headers['x-source-location'] || '';
```
trieloff:
```suggestion
    const sourceLocation = resp.headers['x-source-location'] || '';
    context.sources = [...context.sources, sourceLocation || url.href ];
```

We should store the source, so that it can be included in the Surrogate-Keys header here: https://github.com/adobe/helix-pipeline/blob/87fa02657623338940060296565b8ed1f6ad2a81/src/html/set-surrogate-key.js

src/html/fetch-fstab.js
```diff
@@ -0,0 +1,40 @@
...
+
+  try {
+    // schedule fetch task
+    downloadManager.fetchGithub({
+      owner,
+      repo,
+      ref,
+      path: '/fstab.yaml',
+      id: 'fstab',
+    });
```
trieloff: I'd say, we should store the fstab URL in `context.sources`, too.

src/html/fetch-markdown.js
```diff
@@ -9,137 +9,38 @@
...
+      repo,
+      ref,
+      path,
+      errorOn404: true,
+    });
+
+    content.body = res.body;
 
-  return fetchMarkdown(context, action);
+    // Update sources, but use the branch instead of the resolved ref. todo: why ?
```
trieloff: The `context.sources` will be evaluated by https://github.com/adobe/helix-pipeline/blob/87fa02657623338940060296565b8ed1f6ad2a81/src/html/set-surrogate-key.js to set the `Surrogate-Keys` header, which allows selective invalidation.

src/schemas/action.schema.json
```diff
@@ -35,6 +35,10 @@
     "transformer": {
       "type": "object",
       "description": "A VDOMTransformer instance"
+    },
+    "downloadManager": {
```
trieloff: Maybe rename it to `downloader`? Shorter and less enterprisey.

src/utils/DownloadManager.js
```diff
@@ -0,0 +1,241 @@
...
+
+      if (res.status === 404) {
+        if (!opts.errorOn404) {
+          logger.warn(`Could not find file at ${uri}`);
+          res.body = 'not found';
+          return res;
+        }
+        logger.error(`Could not find file at ${uri}`);
+        setdefault(context, 'response', {}).status = 404;
+      } else if (res.status === 502 || res instanceof TimeoutError) {
```
stefan-guggisberg: `fetch` never returns a `TimeoutError`, it rejects with a `TimeoutError`. The respective check should me moved to the `catch` block above.

test/testDownloadManager.js
```diff
@@ -0,0 +1,413 @@
...
+
+  it('fetch fails if no uri', () => {
+    const action = coerce({
+      logger,
+    });
+    const mgr = new DownloadManager({}, action);
+    assert.rejects(async () => mgr.fetch({}), Error('Unknown uri, cannot fetch content'));
+  });
+
+  it('fetchGithub faile if no owner', () => {
```
stefan-guggisberg: typo
```suggestion
  it('fetchGithub fails if no owner', () => {
```

stefan-guggisberg: typo
```suggestion
  it('fetchGithub fails if no owner', () => {
```

test/testPipeWithFstab.js
```diff
@@ -0,0 +1,292 @@
+/*
+ * Copyright 2018 Adobe. All rights reserved.
```
stefan-guggisberg:
```suggestion
 * Copyright 2020 Adobe. All rights reserved.
```

src/html/fetch-external.js
```diff
@@ -0,0 +1,114 @@
...
+  const fstabTask = downloadManager.getTaskById('fstab');
+  if (!fstabTask) {
+    logger.info('unable to fetch from external. no fstab task scheduled.');
+    return;
+  }
+
+  try {
+    const res = await fstabTask;
+    if (res.status !== 200) {
+      logger.info(`unable to fetch fstab.yaml: ${res.status}`);
```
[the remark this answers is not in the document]
tripodsan: I even think this could be a `debug` :-) I would assume the majority of the projects not to have a `fstab.yaml`....

src/html/fetch-markdown.js
tripodsan: yes. but why do we use the `branch` instead of `ref` ? 

but I think I know, since the `ref` source is stable and never changes....

tripodsan:
@trieloff @stefan-guggisberg I applied your suggestions and removed the `request` fallback completely from the `downloader`. now it only uses `helix-fetch` and the tests force `http1`.

trieloff:
Let's get this merged and find out what happens.

tripodsan:
> Let's get this merged and find out what happens.

ok. I'll take care of the helix-pages update...

Issue #620, opened 2020-03-17
Add Epsagon support for render actions


kptdobe:
Render actions (actions created by the pipeline and the `hlx build` process) do not trace anything inside Epsagon. This is missing to have a complete architecture overview.

PR #624, opened 2020-03-17
feat(pipeline): instrument with Epsagon

kptdobe:
PR for #620


src/utils/openwhisk.js
```diff
@@ -167,7 +168,9 @@ async function runPipeline(cont, pipe, actionParams) {
   if (actionParams.__ow_logger && !actionParams.__ow_logger.trace) {
     actionParams.__ow_logger.trace = actionParams.__ow_logger.silly;
   }
-  return logger(logger.trace(runner))(actionParams);
+
+  const epsagonWrap = epsagon(runner, actionParams);
+  return logger(logger.trace(epsagonWrap))(actionParams);
```
tripodsan: you could also use `openwhisk-action-utils.wrap` for simplicity.

kptdobe: I hesitated but I though saving a dependency will be smarter (this is all packaged during `hlx build`).

trieloff: If it doesn't increase the bundle size of an example action by more than 1%, I'd go with @tripodsan's proposal.

package.json
```diff
@@ -68,6 +69,7 @@
     "micromatch": "^4.0.2",
     "object-hash": "^2.0.0",
     "property-information": "^5.2.2",
+    "proxyquire": "^2.1.3",
```
trieloff:
```suggestion
    "proxyquire": "2.1.3",
```

src/utils/openwhisk.js
kptdobe: zipped or not zipped ? :) It adds 3.15% (zipped)!!
But since we are only talking about 100Ko zipped and I like clean code, I will still do it.

tripodsan: > zipped or not zipped ? :) It adds 3.15% (zipped)!!

just the `wrap` or adding `epsagon` ?

kptdobe: `wrap` and its dependencies.

tripodsan: > `wrap` and its dependencies.

hmm...weird.... I have to investigate this....

PR #626, opened 2020-03-18
fix(fstab): ensure automatic fallback to github on 404

### LINKED ISSUE #625: No automatic fallback to github with fstab
from @davidnuescheler:

i just noticed that for helix-pages examples where we had an `fstab` that was mounted to `/` we had a behaviour that the `header.md` and `footer.md` still were pulled from the github repo. which turns out to be great... it seems that change to helix-pages that was deployed very recently has broken that behavior. i am not sure if things are pulled now from the actual google drive or one drive but they are definitely not pulled anymore from the github repo

tripodsan:
fixes #625


trieloff:
Does this do the fallback sequentially (bad) or speculatively (good)?

tripodsan:
> Does this do the fallback sequentially (bad) or speculatively (good)?

currently, bad :-) it simple doesn't set the `context.content`, so the normal `.md` is tried to be fetched. I would like to fix this quickly in order to unblock @davidnuescheler.

let's tackle the concurrent fetching in a separate PR.

Issue #593, opened 2020-02-20
Data Pipelines with Markdown Templates?


trieloff:
As a continuation of the discussion we had around query-backed pipelines (https://github.com/adobe/helix-home/issues/90) I'm trying to write up a mini-spec around an offhand remark @davidnuescheler made along the lines of "I think query-backed pipelines should be data pipelines – and they should use Markdown templates". This is an obvious riff on @dylandepass's original headless CMS POC, so I'd love to hear his feedback.

## Use case

Some content is just not hypertext-shaped, but effectively spreadsheet-shaped, for instance:
- the latest blog posts for a blog's homepage (coming from a search query)
- the navigation categories on a site, managed in a Google Spreadsheet
- dealer locations for a company homepage, managed through Airtable

## Current Solution

In the current solution (see `theblog` for an example), we would fetch the JSON representation of the data on the client-side, then navigate to "the big array", iterate over it and apply HTML templates (or some other templating language) to update the DOM.

This has the drawback of creating additional engineering complexity and runtime complexity, because you now have to deal with two templating systems (HTL on the server-side and whatever you have on the client side)

## Proposed solution

Instead of using a separate pipeline which would come with the problem of selecting the correct extension or selector, we supercharge the normal Markdown to HTML pipeline, so that it can detect some special directives in the Markdown front matter:

```markdown
---
# this can be just the URL of a known data source, the pipeline will know how to fetch the JSON representation
hlx_data: https://docs.google.com/spreadsheets/d/1IX0g5P74QnHPR3GW1AMCdTk_-m954A-FKZRT2uOZY7k/edit#gid=0
# TDB, we might want to use a different templating language
hlx_template: handlebars
# a JSON pointer expression that points to the root node to start iterating
hlx_root: /sheets[1]
# a sift https://www.npmjs.com/package/sift expression to filter from the root
hlx_filter: 
  maker:
    $eq: Tesla
---

# A list of cars
  {{#each row}}
## {{this.maker}} {{this.model}}

{{this.model}} was first released in {{this.year}}
  {{/each}}
```

Upon detecting a `hlx_data` tag in the front matter, the pipeline will:
1. fetch the associated JSON
2. (maybe post-process it – possible pre.js extension point)
3. use the JSON Pointer from `hlx_root` to navigate to an array node
4. use `sift` and `hlx_filter` to remove elements from the array that don't match
[...]

PR #628, opened 2020-03-18
Revert "Data Embeds"

trieloff:
Reverts adobe/helix-pipeline#611

@rofe reminded me that we still need to discuss the `{{var}}` vs `${var}` vs other syntaxes.


rofe:
I propose to reopen #593 to continue the discussion.

PR #646, opened 2020-03-26
fix(external): make external HTTP timeout configurable

### LINKED ISSUE #638: default timeout of 10s too low for external sources
**Is your feature request related to a problem? Please describe.**
the default timeout to fetch external sources, like word2md is 10s. which is often too slow, especially for cold starts.

**Describe the solution you'd like**
set the default to 20s and make it configurable.

tripodsan:
fixes #638


src/html/fetch-external.js
```diff
@@ -13,15 +13,17 @@ const crypto = require('crypto');
...
+}) {
   setdefault(context, 'content', {});
   // check for fstab
   const fstabTask = downloader.getTaskById('fstab');
   if (!fstabTask) {
     logger.info('unable to fetch from external. no fstab task scheduled.');
     return;
   }
-
+  const timeout = secrets.HTTP_TIMEOUT_EXTERNAL || 20000;
```
trieloff: You don't really need the default here and I'd avoid the duplication.

tripodsan: ok

PR #663, opened 2020-04-09
fix(embed): fix embed url construction

kptdobe:
PR for https://github.com/adobe/helix-pipeline/issues/662

### DIFF
--- a/src/utils/embed-handler.js
+++ b/src/utils/embed-handler.js
@@ -20,7 +20,7 @@ function embed({ EMBED_SERVICE } = {}) {
     const { url } = node;
     let src = url;
     if (URI.parse(url).reference === 'absolute') {
-      src = URI.normalize(`${EMBED_SERVICE}/${url}`);
+      src = URI.normalize([EMBED_SERVICE, EMBED_SERVICE.slice(-1) !== '/' ? '/' : '', url].join(''));
     }
     const props = {
       // prepend the embed service for absolute URLs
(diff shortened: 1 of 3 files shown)


kptdobe:
The smoke tests fails because of https://github.com/adobe/helix-cli/pull/1359
We'll have to merge with failing smoke tests, release and re-run tests from https://github.com/adobe/helix-cli/pull/1359 to make sure this all fits together.

src/utils/embed-handler.js
```diff
@@ -20,7 +20,7 @@ function embed({ EMBED_SERVICE } = {}) {
     const { url } = node;
     let src = url;
     if (URI.parse(url).reference === 'absolute') {
-      src = URI.normalize([EMBED_SERVICE, url.slice(-1) === '/' ? '/' : '', url].join(''));
+      src = URI.normalize([EMBED_SERVICE, EMBED_SERVICE.slice(-1) !== '/' ? '/' : '', url].join(''));
```
tripodsan, line 23: why not formulate it positive? 

```suggestion
      src = URI.normalize([EMBED_SERVICE, EMBED_SERVICE.slice(-1) === '/' ? '' : '/'', url].join(''));
```

kptdobe: It feels more logical to me to read: `if there is no slash, add one, otherwise do nothing` than `if there is a slash, do nothing, otherwise add one`...

PR #679, opened 2020-04-17
feat(all pipes): add clearPositions step to the default pipelines

### LINKED ISSUE #616: Add clearPositions step to the default pipelines
> do we really need positions? I think it's not very useful, even for debugging. maybe we should disable them them completely when generating the mdast. this would also save memory and speed up tree generation, I think.

_Originally posted by @tripodsan in https://github.com/adobe/helix-pipeline/pull/611_

Good idea, we should remove the positions when we are in production mode, maybe even when the log level is greater than `debug`.

ramboz:
Requires #685.

- Add `clearPositions` step right after frontmatter as it is the last step using the info
- Remove the position info from the MDAST if log level is lower than `debug`
- Add unit tests

fix #616


src/defaults/html.pipe.js
```diff
@@ -71,6 +72,7 @@ const htmlpipe = (cont, context, action) => {
     .use(fetch).expose('fetch').when(hascontent)
     .use(parse).expose('parse')
     .use(parseFrontmatter)
+    .use(clearPositions)
```
trieloff, line 75: The conditional should be here.

src/utils/clear-mdast-positions.js
```diff
@@ -0,0 +1,24 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+const removePosition = require('unist-util-remove-position');
+const { numericLogLevel } = require('@adobe/helix-log');
+
+async function clearPositions({ content: { mdast = {} } }, { logger }) {
+  if (numericLogLevel(logger.level) < numericLogLevel('debug')) {
```
trieloff, line 17: It's generally (and in this case, too) a good practice to have the conditionals in the pipeline, and not in the step. This makes your step unit tests easier (no branching needed) and separates the conditional from the working code.

src/defaults/html.pipe.js
trieloff:
```suggestion
    .use(clearPositions).when(production)
```

would be a simple fix, re-using the "are we in production or debug mode" check that already exists.

tripodsan: why not before `parseFrontmatter` ?

trieloff:
@tripodsan integrating `parse-frontmatter` into `remark-parser` is unlikely to happen, especially with our extensions and beyond the scope of this pull request. I think that @ramboz approach is correct in general, just needs a minor tweak in detail.

tripodsan:
> @tripodsan integrating `parse-frontmatter` into `remark-parser` is unlikely to happen, especially with our extensions and beyond the scope of this pull request.

why? it shouldn't be too difficult to extend the parser and might be quite cool. (and I know that @ramboz like to do cool things)

> I think that @ramboz approach is correct in general, just needs a minor tweak in detail.

I think avoiding the `position` object would save some overhead, so better not generate it at all.

trieloff:
I think it's unlikely to happen because there is already https://github.com/remarkjs/remark-frontmatter, so this will probably not make it into the core.

We've been using the remark-frontmatter plugin before and then switched away from it. We might want to switch back, and while I can't comment on the coolness of that, I'd definitely see this as being a lot of work for a relatively minor benefit.

We can certainly discuss this, but I'd rather have it separate from this PR, which is about two orders of magnitude smaller.

tripodsan:
> I think it's unlikely to happen because there is already https://github.com/remarkjs/remark-frontmatter, so this will probably not make it into the core.

As I said, not trying to contribute/update the existing parser or plugin, but to write our own.

tripodsan:
@trieloff ... it wasn't that difficult: https://github.com/adobe/helix-pipeline/pull/685 ;-)

ramboz:
@tripodsan 
> the only thing that needs them is the parse-frontmatter, which would probably be better incorporated right into the remark-parser anyways...

Yes, I just didn't go that route because I didn't want to touch the frontmatter step, but since you are addressing it in #685, I'll switch to completely disabling positions instead of removing them. I'll also handle the toggle between debug and normal mode

src/html/parse-markdown.js
```diff
@@ -12,18 +12,22 @@
...
 
   const request = setdefault(context, 'request', {});
   if (!request.extension) request.extension = 'html';
   const { extension } = request;
 
-  action.logger.debug(`Parsing markdown from request body starting with ${body.split('\n')[0]}`);
-  content.mdast = unified().use(remark).parse(body);
+  const isDebugLevel = numericLogLevel(logger.level) >= numericLogLevel('debug');
+  logger.debug(`Parsing markdown from request body starting with ${body.split('\n')[0]} with positions ${isDebugLevel ? 'enabled' : 'disabled'}`);
+  content.mdast = unified().use(remark).use({ settings: { position: isDebugLevel } }).parse(body);
```
tripodsan, line 29: simpler ?

```suggestion
  content.mdast = unified().use(remark, { position: isDebugLevel }).parse(body);
```

tripodsan:
@ramboz I'm going to merge this, once the build passes. ok ?

Issue #683, opened 2020-04-21
Change logic of generating docs


tripodsan:
Generate-docs fails on pipeline:
https://circleci.com/gh/adobe/helix-pipeline/6214?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification (edited) 

due to the changed protected branch policies.

```
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: 7 of 7 required status checks are expected. At least 2 approving reviews are required by reviewers with write access.  
```

I think that generating docs should not be a CI task after the release, but should happen in the PR before a release. otherwise the docs don't correlate with the release tag

### suggestion

- add run docs to lint
- add step to ci that runs the docs and then checks if git is still clean

PR #684, opened 2020-04-21
Ensure that docs are up-to-date in CI

tripodsan:
see #683 

- changed `generate-docs` to only run on non-master, non-renovate branches
- `generate-docs` build the docs and halts the step if nothing is modified
- otherwise it will commit the the docs.

example CI when nothing is modified:
- https://circleci.com/gh/adobe/helix-pipeline/6250

example commit that triggered new docs:
- schema update: 0bd5144d2ca4339eac098c68cc229b26f056dc50
- ci workflow: https://circleci.com/gh/adobe/helix-pipeline/6253
- docs commit by ci: d468956a85eedcc769e99b1204533f31a53d2a92


package.json
```diff
@@ -12,7 +12,7 @@
     "lint": "npx eslint . && npm run types",
     "types": "npm run docs && node bin/schema2ts.js",
     "docs": "npx jsonschema2md -d src/schemas -o docs -x docs",
-    "commit": "git-cz"
+    "commit": "npm run types && git-cz"
```
trieloff, line 15: This isn't super useful because the generated docs won't be `git add`ed.

tripodsan: but it would generate the docs and then the dev sees if they are outdated

tripodsan:
> I'm a bit ambivalent on this. Not pushing directly to master is definitely an improvement, but putting additional burden on the developer is only leading to frustration and outdated documentation.

I don't see it like this, as the developer anyways should run `npm run lint` before comitting.
This will already generate the docs.

> Wouldn't pushing the documentation into the branch work?

yes...this is probably better...

.circleci/config.yml
```diff
@@ -15,8 +15,8 @@ commands:
           keys:
             - v2-dependencies-{{ arch }}-{{ checksum "package-lock.json" }}
       - run:
-          name: install npm 6.9.0
-          command: sudo npm -g install npm@6.9.0
+          name: install latest npm
+          command: sudo npm -g install npm
```
rofe, line 19: Why change this?

trieloff:
The tidelift issue seems like a false positive. I've opened a support ticket.

.circleci/config.yml
tripodsan: because I'll change it everywhere? I think we should always use the latest npm.
since npm shows a nice banner when it's outdated, I think that users tend to do it.
also by using the latest npm, we might discover problems with new versions ealry.

PR #685, opened 2020-04-21
feat(md): create frontmatter plugin

### LINKED ISSUE #616: Add clearPositions step to the default pipelines

### LINKED ISSUE #697: Resilience: FrontMatter is too strict
I was authoring a markdown file (stored in onedrive but I think this does not really matter) and I somehow forgot some line breaks around the sections breaks:

```
# Multimedia Test
---
![](https://hlx.blob.core.windows.net/external/20f9d6dff67514da262230822bda5f3b50ef28c6#image.png)
---
PUBLISHED ON 28-04-2020
---

### SoundCloud

<https://soundcloud.com/adobeexperiencecloud/experience-makers-in-travel-and-hospitality>

### YouTube

<https://www.youtube.com/embed/L8YPmZ6jYyQ?rel=0>

### SlideShare

<https://www.slideshare.net/adobe/adobe-digital-insights-holiday-recap-2019>

---
Topics: Bar, Baz
Products: Stock, Creative Cloud
```

This fires a 500 (via helix-pages):

```
Exception during #05/use:parseFrontmatter from /nodejsAction/XXKFHkjA/html.js:225207: TypeError: Cannot read property 'idx' of null at map (/nodejsAction/XXKFHkjA/html.js:257850:17) at map (/nodejsAction/XXKFHkjA/html.js:4047:11) at map.next (<anonymous>) at filter (/nodejsAction/XXKFHkjA/html.js:4065:14) at filter.next (<anonymous>) at flat (/nodejsAction/XXKFHkjA/html.js:4296:16) at flat.next (<anonymous>) at trySlidingWindow (/nodejsAction/XXKFHkjA/html.js:4552:14) at trySlidingWindow.next (<anonymous>) at filter (/nodejsAction/XXKFHkjA/html.js:4065:14)
```

This also triggered an escalation because of too many 5xx errors.
Adding some few line breaks here and there fixed the issue.

This is not a good user-experience and I am sure there is a way to pre-process the md before the mdast transformation to add the missing required line breaks for me and avoid those errors while authoring.

tripodsan:
- fixes #616
- fixes #697
- alternative to #679


src/html/parse-markdown.js
```diff
@@ -23,7 +24,12 @@ function parseMarkdown(context, action) {
   const { extension } = request;
 
   action.logger.debug(`Parsing markdown from request body starting with ${body.split('\n')[0]}`);
-  content.mdast = unified().use(remark).parse(body);
+  content.mdast = unified()
+    .use(remark, {
+      position: false,
```
ramboz, line 29: 👍 That essentially replaces my other PR. Now if we want to keep positions in there in case of debug logs, we may need to make this configurable. I'd propose to not address the position removal here and keep it in the other PR for clarity

trieloff:
I like that it's not reporting an error.

src/html/parse-frontmatter.js
```diff
@@ -136,188 +129,85 @@ const production = require('../utils/is-production');
...
-};
+function plugin(opts) {
+  const { blockMethods, blockTokenizers } = this.Parser.prototype;
+  blockMethods.unshift('frontmatter');
+  blockTokenizers.frontmatter = create(opts);
+}
 
-assign(parseFrontmatter, { findFrontmatter, FrontmatterParsingError });
-module.exports = parseFrontmatter;
+module.exports = plugin;
```
trieloff, line 212: This needs to become a breaking change. If someone is constructing their own pipeline, and imports the `parseFrontmatter` step, it will break. Alternative: export a `nop` function that can still be imported, but that does nothing.

tripodsan: yes...I thought of that, too.

tripodsan: let's do both...

src/html/parse-frontmatter.js
```diff
@@ -9,315 +9,27 @@
...
-      && next.type === 'frontmatter'
-      && val.end === next.start),
-    /* eslint-disable-next-line no-unused-vars */
-    map(([val, next]) => val),
-  );
-};
+/**
+ * @deprecated
+ */
+const findFrontmatter = () => [];
```
trieloff, line 19: Why an empty array? Wouldn't `undefined` work as well?

tripodsan: AFAICS, the original `findFrontmatter()` returned an array.

PR #686, opened 2020-04-21
test(embeds): add test examples from word2md and gdocs2md for embed t…

trieloff:
…esting

/cc @stefan-guggisberg


trieloff:
I can't approve it, although I think it's an awesome PR.

Issue #664, opened 2020-04-10
investigate why requests are not concurrent


tripodsan:
![image](https://user-images.githubusercontent.com/917628/78961336-120f6e00-7b2c-11ea-8658-92116dc8b534.png)

why is the 2nd request to raw.github (`helix-markup.yaml`) AFTER the word2md? 
It should be parallel to fetching the fstab.yaml

trieloff:
https://github.com/adobe/helix-pipeline/blob/4489463b5e0a2d36e8c770c74d8da71a6a006826/src/defaults/html.pipe.js#L68-L70

If we swap `fetchExternal` and `fetchMarkupConfig`, it should work in parallel, as `fetchFstab` and `fetchMarkupConfig` are non-blocking, but `fetchExternal` is blocking.

stefan-guggisberg:
@tripodsan Can this issue be closed?

tripodsan:
yup.

PR #691, opened 2020-04-24
feat(html pipe): Content Intelligence Matching from markup.yaml

### LINKED ISSUE #687: Content Intelligence Matching from markup.yaml
With https://github.com/adobe/helix-shared/pull/280 in the books, we can start on https://github.com/adobe/helix-shared/issues/278

@ramboz let me know if this is something for you.

ramboz:
Add support for content intelligence matching in the helix-markup.yaml to patch the MDAST

fix #687, #664


src/utils/adjust-markup.js
```diff
@@ -85,36 +143,32 @@ async function adjustMDAST(context, action) {
...
 
+  // Adjust the MDAST based on AST matcher
   Object.entries(markupconfig.markup)
     .filter(([_, cfg]) => cfg.type === 'markdown')
     .forEach(([name, cfg]) => {
       logger.info(`Applying markdown markup adjustment: ${name}`);
-
       transformer.match(cfg.match, (h, node) => {
         const handler = transformer.constructor.default(node);
-        let el = handler(h, node);
```
ramboz, line 95: This is just extracted to `patchHastNode` above

src/utils/adjust-markup.js
```diff
@@ -136,26 +190,8 @@ async function adjustHTML(context, { logger, markupconfig }) {
     .filter(([_, cfg]) => !cfg.type || cfg.type === 'html')
     .forEach(([name, cfg]) => {
       logger.info(`Applying HTML markup adjustment: ${name}`);
-
       const elements = context.content.document.querySelectorAll(cfg.match);
-      elements.forEach((el) => {
```
ramboz, line 141: This is just extracted to `patchHtmlElement` above

test/testPipeWithMarkupConfig.js
```diff
@@ -52,20 +76,8 @@ describe('Testing HTML Pipeline with markup config', () => {
...
-  it('html.pipe adjusts the MDAST as per markup config', async () => {
-    nock('https://raw.githubusercontent.com')
-      .get('/adobe/test-repo/master/helix-markup.yaml')
-      .reply(() => [200, TEST_MARKUP_CONFIG_MD])
-      .get('/adobe/test-repo/master/hello.md')
-      .reply(() => [200, '# Hello\nfrom github.\n']);
-
-    const context = {};
-    const action = coerce({
+    context = {};
```
ramboz, line 79: Just extracting `context` and `action` into the `beforeEach`

test/testPipeWithMarkupConfig.js
```diff
@@ -39,8 +39,32 @@ markup:
...
+      baz: qux
+    wrap: .corge
+    type: content
+`;
+
+function cleanup(html) {
+  return html.replace(/\n\s*/g, '').trim();
+}
+
+function expectBodyEquals(result, expectedMarkup) {
```
ramboz, line 58: Create util method to reduce boilerplate code

test/testPipeWithMarkupConfig.js
```diff
@@ -81,19 +93,75 @@ describe('Testing HTML Pipeline with markup config', () => {
...
+          <h1 id="hello">Hello</h1>
+          <p>from github.</p>
+        </div>
+      </div>
+      <div>
+        <h1 id="bar">Bar</h1>
+      </div>`);
+  });
+
+  it('html.pipe does not adjust the MDAST as per markup content config if there are no sections', async () => {
```
trieloff, line 148: If there are no sections, I’d assume we treat the document as one big section.

tripodsan:
@ramboz I'm not sure that i 100% understand how this works, but is it operating on the vdom or the hast? (since there is a `patchHastNode()`, I assume the later).

you could add some comments to the file that explains what is happening here.

ramboz:
@tripodsan For the MDAST part, we are getting an MDAST node as input and are hijacking the HAST transformation to patch the output node. So the selector applies to the MDAST tree, but the properties are added to the output HAST node. Does that make sense?

I'll add better comments to clarify

src/defaults/html.pipe.js
```diff
@@ -65,8 +65,8 @@ const htmlpipe = (cont, context, action) => {
     .every(validate).when((ctx) => !production() && !ctx.error)
     .every(timer.update)
     .use(fetchFstab)
-    .use(fetchExternal)
     .use(fetchMarkupConfig)
+    .use(fetchExternal)
```
ramboz, line 69: This should fix #664 at the same time

tripodsan:
> So the selector applies to the MDAST tree, but the properties are added to the output HAST node. Does that make sense?

but we are not using the HAST anymore, only the vdom....

ramboz:
@tripodsan You are obviously right… I got confused between HAST and VDOM.
Updated accordingly

Issue #702, opened 2020-05-06
Support Vimeo embeds


kptdobe:
See https://theblog.adobe.com/slow-down-and-speed-up-clips-in-premiere-rush/.

Required for https://github.com/adobe/theblog/issues/186

kptdobe:
> I couldn't find the exact video URLs

I think that's a general issue, it seems that you can export the video and only get the player url. For Soundcloud, I had the same issue that I solved by requesting the player url and extract the direct url from the dom meta (see https://github.com/adobe/helix-theblog-importer/blob/master/src/index.js#L44-L64)

Video does not provide the "direct" url. See https://player.vimeo.com/video/349091463

One option would be to keep the player url and `helix-embed` would be able to deal with it.
WDYT ?

trieloff:
I created a private video (only visible for people who have the URL), copied the URL in Vimeo and it still works: https://adobeioruntime.net/api/v1/web/helix/helix-services/embed@v1/https://vimeo.com/415498678/5349aa3e1f – so for private videos I think authors will be able to still share and embed them.

kptdobe:
For the import, it is more the other way around: we only have the player url, I can use it for the embed but then it does not work: https://adobeioruntime.net/api/v1/web/helix/helix-services/embed@v1/https://player.vimeo.com/video/349091463

kptdobe:
My proposal is to make helix-embed smarter and use the player url as is if this is the one provided.

MarquiseRosier:
{
        "s": "^https?:\\/\\/player\\.vimeo\\.com\\/video\\/(\\d+)",
        "m": "i"
    },

This is on the list actually!

Issue #701, opened 2020-05-06
Support Instagram embeds


kptdobe:
See https://theblog.adobe.com/wrapping-feelings-in-works-of-art-with-photoshop-on-the-ipad/

Needed for https://github.com/adobe/theblog/issues/186

Issue #700, opened 2020-05-06
Support Twitter embeds


kptdobe:
See https://theblog.adobe.com/the-best-of-summit-2020-so-far/.

Required for https://github.com/adobe/theblog/issues/186

kptdobe:
oh yeah, this would have been pretty unexpected ;)

Good point, I did not check if this is already supported :)
At least, no fully.

PR #703, opened 2020-05-06
Whitelist more embeds

trieloff:
Fix for #700 #701 and #702 

For adobe/theblog#186


MarquiseRosier:
I think we should just get the list from IFramely and auto generate the whitelist so that we don't run into the issue of not supporting embeds unless it's something we really don't support yet!

trieloff:
I think using the iFramely whitelist is a good idea for helix pages. Let’s open another issue for that.

MarquiseRosier:
> I think using the iFramely whitelist is a good idea for helix pages. Let’s open another issue for that.

Ok! :)

PR #728, opened 2020-05-22
fix(all pipes): get-metadata is not map-enumeration stable

### LINKED ISSUE #669: get-metadata is not map-enumeration stable
**Description**
the `meta` properties on the mdast nodes are computed based on the entries of an object:

```js
    types.push(...Object.entries(typecounter) // get pairs of type, count
      .sort((left, right) => left[1] < right[1]) // sort descending by count
      .slice(0, 3) // take the top three
      .map(([name]) => name) // keep only the type
      .reduce((names, name) => [`${names[0] || 'is'}-${name}`, ...names], [])); // generate names
```

this leads to different results, depending on the enumeration.

**To Reproduce**
- run helix-cli test on node10 vs node 12:

```
      + expected - actual

             "nb-text-3"
             "nb-blockquote-1"
             "nb-delete-1"
             "nb-heading-1"
      -      "is-html-text-blockquote"
      -      "is-html-text"
      -      "is-html"
      +      "is-text-html-blockquote"
      +      "is-text-html"
      +      "is-text"
           ]
         }
```

ramboz:
Force alphabetical sorting of types in case of collisions.

fix #669


trieloff:
Leave @lkrapf out of this, you need the *other* Lars ;)

ramboz:
Blocked by https://github.com/adobe/helix-cli/pull/1412

PR #737, opened 2020-05-28
feat(pipe): move image handler into pipeline step

### LINKED ISSUE #706: Move image handler into pipeline step before "meta"
Helix Pages now uses `context.content.image` in the metadata, but in case of Azure blob store images it unfortunately doesn't contain the `/hlx_*` URL.

@tripodsan suggested to move the image handler, which is currently rewriting blob store links on the VDOM as part of the "html" step, into its own "image" step and insert it before "meta", so the image URLs are already rewritten in the mdast when metadata is being extracted.

rofe:
Fix #706


src/html/rewrite-blob-images.js
```diff
@@ -0,0 +1,34 @@
...
+function rewrite({ content }) {
+  if (!content.mdast) {
+    return;
+  }
+  visit(content.mdast, (node) => {
+    if (node.type === 'image'
+      && /^https:\/\/hlx\.blob\.core\.windows\.net\/external\//.test(node.url)) {
+      const { pathname, hash } = new URL(node.url);
+      const filename = pathname.split('/').pop();
+      const extension = hash.includes('.') ? hash.split('.').pop() : 'jpg';
```
trieloff, line 28: Maybe we need this: https://github.com/adobe/helix-shared/issues/301

PR #743, opened 2020-06-03
feat(pipe): use content-proxy to download content

### LINKED ISSUE #734: Fetch Markdown from Helix Content Proxy
Paraphrasing @davidnuescheler 

> GitHub is down more often than Fastly, so we should put a cache in front of it. Ideally one like `helix-content-proxy` that can also take care of the fstab resolution and fetch from the correct location, be it GitHub, Google, Word, or whatever else we come up with. 

This would enable us to have a more solid cache for content that is managed by the customer's Fastly instance and can be cleared on demand. Theoretically, we would just replace the `.html` in the original request URI with `.md` (https://github.com/adobe/helix-publish/issues/452) to get the Markdown.

### Challenges

As outlined in https://github.com/adobe/helix-publish/issues/452 we need to agree on an approach that makes serving raw Markdown with long cache lifetimes practical from Fastly.

tripodsan:
fixes #734

BREAKING CHANGE: removed `fetch-fstab` and `fetch-external`


rofe:
Why is this considered a breaking change? As far as I can see, you are only adding a new step and extension point, not changing, removing or reordering extisting ones.

tripodsan:
> Why is this considered a breaking change? As far as I can see, you are only adding a new step and extension point, not changing, removing or reordering extisting ones.

because I removed the `fetch-fstab` and `fetch-external` which might be used directly in a pre.js.

tripodsan:
@trieloff I kept a fallback to load from github, if the github coordinates look like the ones from helix-simulator.

I don't like this too much. As an alternative we could make the content-proxy URL configurable (via secrets) and then tweak it in helix-simulator accordingly.

I think, once we have the content-proxy served via fastly, we anyways need to pass in the respective url.

src/html/fetch-content.js
```diff
@@ -0,0 +1,77 @@
...
+}) {
+  setdefault(context, 'content', {});
+  /* istanbul ignore next */
+  if (!request || !request.params) {
+    throw new Error('Request parameters missing');
+  }
+  const { content } = context;
+  const {
+    owner, repo, path, branch = 'master', ref,
+    contentProxyUrl = 'https://adobeioruntime.net/api/v1/web/helix/helix-services/content-proxy@v1',
```
tripodsan, line 25: @trieloff this would be set by VCL to use the correct runtime namespace. we want to run the content-proxy in the customers namespace.

tripodsan:
The smoke tests fail, because the polly recordings in helix-cli still use the raw.github to fetch the content.

src/defaults/json.pipe.js
```diff
@@ -29,6 +28,10 @@ const timing = require('../utils/timing');
 
 /* eslint newline-per-chained-call: off */
 
+function hasNoContent({ content }) {
+  return !(content !== undefined && content.body !== undefined);
```
kptdobe, line 32: `content.body` could also be null, no ?

why not simply:

```suggestion
  return content && content.body;
```

src/html/fetch-content.js
trieloff: I'd take the namespace from `__ow` so that we don't need VCL changes.

tripodsan: hmmm....good point :-)

src/defaults/json.pipe.js
tripodsan: so that you can specify a `''` content.

tripodsan: > `content.body` could also be null, no ?

I guess not.

PR #761, opened 2020-06-15
fix: don't ignore content proxy

tripodsan:


tripodsan:
**note**: the smoke tests fail, because the simulator isn't sending the correct content-proxy parameter yet. this is a chicken-and-egg problem with https://github.com/adobe/helix-simulator/pull/452

PR #768, opened 2020-06-16
Source hash

tripodsan:
re-adds the `content.data.sourcehash` that was lost during content-proxy replacement

### DIFF
--- a/src/html/fetch-content.js
+++ b/src/html/fetch-content.js
@@ -9,6 +9,7 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
+const crypto = require('crypto');
 const { setdefault } = require('ferrum');
 
 async function fetchContent(context, {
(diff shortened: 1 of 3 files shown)


src/html/fetch-content.js
```diff
@@ -64,8 +65,17 @@ async function fetchContent(context, {
     }
 
     content.body = res.body;
+    setdefault(content, 'sources', []).push(downloader.computeGithubURI(owner, repo, branch, path));
 
-    setdefault(context.content, 'sources', []).push(downloader.computeGithubURI(owner, repo, branch, path));
+    // store extra source location if present
+    const sourceLocation = res.headers.get('x-source-location') || '';
+    const sourceHash = crypto.createHash('sha1').update(sourceLocation).digest('base64').substring(0, 16);
```
trieloff, line 72: Isn't there a helper in `helix-shared`?

tripodsan: no. this is internal to pipeline. it's not used elsewhere.

PR #780, opened 2020-06-30
chore(ci): remove auto-docs workflow

tripodsan:


trieloff:
Why do you want to remove it?

tripodsan:
> Why do you want to remove it?

see the checks in https://github.com/adobe/helix-pipeline/pull/779

the `[skip ci]` confused the checks and the coverage checks remain yellow. we can also remove the `[skip ci]` and see what happens :-) is probably better

PR #789, opened 2020-07-06
fix: use x-source-location from content-proxy for surrogate computation

tripodsan:


src/html/fetch-content.js
```diff
@@ -22,8 +22,9 @@ async function fetchContent(context, {
   }
   const { content } = context;
   const {
-    owner, repo, path, branch = 'master', ref,
+    owner, repo, path, ref,
   } = request.params;
+  const branchOrRef = request.params.branch || ref || 'master';
```
trieloff, line 27: Are you sure about `request.params.branch`?

tripodsan: pretty much... the schema says: `branch` defaults to `ref` defaults to `master`.
so if a `branch` is provided, it is preferred.

without this, this test failed: https://github.com/adobe/helix-pipeline/blob/master/test/testHTML.js#L646


IMO, we should get rid of `branch` and just use `ref` everywhere.

trieloff: I think the reason we introduced `branch` was to enable invalidations. And that means that the new source URL is not sufficient.

If branch `publish` has a current ref `…2cba` and gets a content update, then we need to invalidate hashes that include `publish`, but hashes for `…2cba` can stay the same.

So I'd say this PR is probably wrong.

tripodsan: ok. I fixed it.

tripodsan:
> An issue to explain why we need to use ref or branch for surrogate computation would help me understanding the change...

this is basically a regression fix due to the changes using the content-proxy.

- in order to fetch the most stable content, we prefer `ref` over `branch`, as it contains the SHA of the commit.
- in order to compute the surrogate-key for cache flushing, the `ref` is of little use, as the SHA makes the resource immutable. here we prefer the `branch` over the `ref`, as we can actually use it to flush the cache for a resource on a specific branch.

PR #837, opened 2020-09-01
feat(icons): implement :#notation: to address icons in a collection

### LINKED ISSUE #817: Icons currently not working
From @davidnuescheler in #509:
> currently we seem to render something like:
> 
> ```
> <svg xmlns="http://www.w3.org/2000/svg" class="icon icon-twitter">
> <use href="/icons/twitter.svg"></use>
> </svg>
> ```
> 
> which simply doesn't work without the id and `#` in the `href`.
> 
> so i would propose that we switch to the `icons.svg#<id>` notation unless someone has a `use` version that works without touching the actual svg file.

> after looking a little bit more into this and having usecases where it is pretty useful to use the same icons also for the favicons, i think that a project should be able to support both. icons in the `/icons` folder and icons that are in the `icons.svg` symbol sheet.
> 
> the easiest way to distinguish those would be to support `:logo:` and `:#logo:` notation respectively...
> 
> `:logo:` would product something like a `` while `:#logo:` would produce `<svg class="icon icon-logo"><use xlink:href="/icons.svg#logo"></use></svg>`

rofe:
BREAKING CHANGE: svg icons are now rendered differently

Fix #817


rofe:
To test locally:
1. check out a helix pages project
2. `hlx up`, when started hit `^C`
3. `cd .hlx/pages/master/node_modules/@adobe`
4. `rm -rf ./helix-pipeline`
5. `git clone --single-branch --branch fix-icons https://github.com/adobe/helix-pipeline.git`
6. `cd ../../../..`
7. `hlx up`

trieloff (approved):

kptdobe (approved):

PR #839, opened 2020-09-02
feat(markup): allow variables in markup config

trieloff:
By using `wrap: pre[data-embed="\${url}"]` you can now wrap the HTML with a pre tag that will have a `data-embed` attribute taken from the `url` attribute of the matched MDAST node. 

Also fixes an issue with markup mappings for custom MDAST nodes such as `embed`. 

Also fixes an issue with markup mappings for MDAST nodes that would produce more than one DOM node such as the `esi:include` `esi:remove` pair from `embed`

useful for https://github.com/adobe/helix-home/issues/148


src/utils/adjust-markup.js
```diff
@@ -168,11 +205,35 @@ async function adjustMDAST(context, action) {
...
-        const handler = transformer.constructor.default(node);
+      transformer.match(cfg.match, function myhandler(h, node, parent) {
+        const [fallbackhandler] = transformer
+          .allmatches(node)
+          .filter((theirhandler) => theirhandler !== myhandler);
         // Generate the matching VDOM element
-        const el = handler(h, node);
-        return patchVDOMNode(el, cfg);
+
+        /**
```
ramboz, line 214: Slight indentation bug

ramboz (approved):
Looking good!

tripodsan:
It's hard to judge from looking at the changes...but what's the performance / memory impact ?

src/utils/adjust-markup.js
```diff
@@ -12,17 +12,29 @@
...
+const { get } = require('dot-prop');
 const { match } = require('./pattern-compiler');
 const section = require('./section-handler');
+const VDOMTransformer = require('./mdast-to-vdom');
 
-/** Pleceholder variable for the generate template. */
+/** Placeholder variable for the generate template. */
 const PLACEHOLDER_TEMPLATE = /\$\{\d+\}/g;
 
+function findAndReplace(root, insert) {
```
trieloff, line 28: @tripodsan my own implementation is almost the same as `"hast-util-find-and-replace"`, but with fewer special cases we don't need, so likely a net positive.

src/utils/adjust-markup.js
```diff
@@ -59,15 +71,20 @@ function isSection(node) {
   return node.type === 'root' || node.type === 'section';
 }
 
+function populate(template, data) {
```
trieloff, line 74: Super cheap stuff here.

src/utils/adjust-markup.js
```diff
@@ -168,11 +205,35 @@ async function adjustMDAST(context, action) {
...
+        /**
+     * A function that enables the recursive processing of MDAST child nodes
+     * in handler functions.
+     * @param {function} callback the HAST-constructing callback function
+     * @param {Node} childnode the MDAST child node that should be handled
+     * @param {Node} mdastparent the MDAST parent node, usually the current MDAST node
+     * processed by the handler function
+     * @param {*} hastparent the HAST parent node that the transformed child will be appended to
+     */
+        function handlechild(callback, childnode, mdastparent, hastparent) {
```
trieloff, line 223: This has the biggest potential impact (because it recurses), but it's part of the bugfix and the trees are relatively small.

trieloff:
@tripodsan @ramboz style fixed and back to 100% coverage

tripodsan (approved):

kptdobe (approved):

PR #844, opened 2020-09-04
feat(package): switch to exact dependency versions

### LINKED ISSUE #843: Switch to exact dependency versions
Follow up of https://github.com/adobe/theblog/issues/416

This module depends on a lot of tiny dependencies, especially in the mdast / hast / dom / md transformations and manipulations. Those dependencies are referenced with the range operator which makes it hard to know which one of them breaks something or has an important impact.

Proposal: switch to exact dependency versions (like everywhere else)

kptdobe:
Fix #843 

- switch to exact dep
- update for each of them
- audit fix on the project (6 security fixes)


package.json
```diff
@@ -58,44 +58,44 @@
...
+    "js-yaml": "3.14.0",
     "jsdom": "16.4.0",
-    "lodash": "^4.17.15",
-    "mdast-util-to-hast": "^9.0.0",
-    "mdast-util-to-string": "^1.0.6",
-    "micromatch": "^4.0.2",
-    "object-hash": "^2.0.0",
-    "path-to-regexp": "^6.1.0",
-    "property-information": "^5.2.2",
+    "lodash": "4.17.20",
```
tripodsan, line 74: do we need lodash? maybe we can replace it with only the package of the function that we need

package.json
```diff
@@ -58,44 +58,44 @@
...
+    "mdast-util-to-hast": "9.1.1",
+    "mdast-util-to-string": "1.1.0",
+    "micromatch": "4.0.2",
+    "object-hash": "2.0.3",
+    "path-to-regexp": "6.1.0",
+    "property-information": "5.5.0",
     "proxyquire": "2.1.3",
```
tripodsan, line 81: why is proxyquire not in the `devDeps` ?

package.json
```diff
@@ -58,44 +58,44 @@
...
-    "request": "^2.87.0",
-    "request-promise-native": "^1.0.7",
-    "retext": "^7.0.0",
-    "retext-smartypants": "^4.0.0",
-    "strip-markdown": "^3.1.0",
-    "unified": "^9.0.0",
-    "unist-util-map": "^2.0.0",
-    "unist-util-remove": "^2.0.0",
+    "remark-parse": "8.0.3",
+    "request": "2.88.2",
```
tripodsan, line 83: do we still need those?

package.json
```diff
@@ -58,44 +58,44 @@
...
+    "request-promise-native": "1.0.9",
```
tripodsan, line 84: do we still need those?

trieloff (approved):

package.json
kptdobe: Moved to `devDeps`

package.json
kptdobe: We need lodash in several places.

package.json
kptdobe: `remark-parse` needed.
`request` is required by `jsdom` (even though it is not documented).
Removed `request-promise-native`.

package.json
kptdobe: Yes, `request` is required by `jsdom`.

trieloff (approved):

trieloff:
let's get a review from @rofe before this goes stale again

tripodsan (approved):

PR #878, opened 2020-10-26
fix(deps): upgrade remark-parse to v9

### LINKED ISSUE #873: migrate remark to v9
see https://github.com/adobe/helix-pipeline/pull/870

### LINKED ISSUE #883: regression escaping tags from gdoc
this may or may not be a pipeline issue.

in the pages project we used markers in the document that look like tags

```
<steps>
```

this used to be rendered to

```
<p>&lt;steps&gt;</p>
```

and is now rendered to

```
<p>\<steps></steps></p>
```

i think this points to a number of issues at hand. first we need to upgrade our test harness and secondly this looks like tags are not escaped anymore coming from a google doc?

trieloff:
fixes #873
fixes #883 

BREAKING CHANGE: using the new remark parser introduces some breaking changes to the mdast.
   - link references that don't have a definition wont' generate a mdast node.
   - images now contain empty string alt attributes if none specified in the markdown 
   see https://github.com/remarkjs/remark/releases/tag/13.0.0 for more details.


test/fixtures/wordembeds.json
```diff
@@ -70,31 +68,33 @@
...
         {
           "type": "text",
-          "value": "And another example with a QR code: "
+          "value": "And another example with a QR code:"
         }
       ]
     },
     {
-      "type": "paragraph",
+      "type": "embed",
```
trieloff, line 76: @tripodsan there are a few `paragraph`s that come out as `embed`s in with r@9 – is that intended?

test/fixtures/embeds.html
```diff
@@ -25,14 +25,14 @@ <h2 id="image-style-embeds">Image-Style Embeds</h2>
     src="https://adobeioruntime.net/api/v1/web/helix/helix-services/embed@v1/https://www.youtube.com/watch?v=KOxbO0EI4MA">
   </esi:include>
   <esi:remove>
-    <p><img src="https://www.youtube.com/watch?v=KOxbO0EI4MA"></p>
+    <p><img src="https://www.youtube.com/watch?v=KOxbO0EI4MA" alt=""></p>
```
tripodsan, line 28: **Note**: The new markdown parser now generates empty string values for the `alt` properties in mdast. they used to be `null`.  I think this doesn't affect the semantics of the HTML.

test/fixtures/googleembeds.json
```diff
@@ -6,7 +6,7 @@
       "children": [
         {
           "type": "text",
-          "value": "Hello, Google Docs Embed! "
+          "value": "Hello, Google Docs Embed!"
```
tripodsan, line 9: Slightly different handling of trailing spaces. shouldn't affect HTML generation.

test/fixtures/googleembeds.json
```diff
@@ -15,38 +15,40 @@
...
         {
           "type": "text",
-          "value": "Here’s an image with a QR Code: The QR Code resolves to the url of a Youtube clip. This image will be replaced by a Link+Image-style Embed. "
+          "value": "Here’s an image with a QR Code: The QR Code resolves to the url of a Youtube clip. This image will be replaced by a Link+Image-style Embed."
         }
       ]
     },
     {
-      "type": "paragraph",
+      "type": "embed",
```
tripodsan, line 23: @trieloff can you take a 2nd look here? don't know if this is still correct.

test/fixtures/googleembeds.json
trieloff: I think it's fine.

test/fixtures/wordembeds.json
tripodsan: don't know :-) what should be the correct behaviour ?

test/fixtures/embeds.html
trieloff: It does affect the semantics, empty `alt` means screen readers will skip the image entirely because it's just decoration, missing `alt` attributes will cause the `src` attribute to be read. There may be an existing issue in remark already.

test/fixtures/googleembeds.json
trieloff: looks iffy: https://github.com/adobe/helix-pipeline/pull/878#discussion_r519642694

test/fixtures/embeds.html
tripodsan: which is actually a good thing:
https://www.w3.org/WAI/tutorials/images/decorative/

without having a default to "", there is no way to create a decorative image in markdown.
I think this is a good behaviour - however, I filed a bug anyways: https://github.com/micromark/micromark/issues/32

test/fixtures/wordembeds.json
tripodsan: so the MD is something like:

```
[![Lars explains the future.](https://hlx.blob.core.windows.net/external/aa59825575016ba5bb9f1ee54be5d7efc23e3032#image.png)](https://www.youtube.com/watch?v=85UKLsvQEIA) 
```

which is correctly detected as `iaEmbed`
maybe this was broken before?

test/fixtures/embeds.html
trieloff: Thanks for filing the issue, I agree with @wooorm's trade-offs.

test/fixtures/googleembeds.json
trieloff: No, it is correct.

test/fixtures/wordembeds.json
trieloff: I think it was either broken before, now it's good.

trieloff (commented):
Looks good now, the embeds were not detected before, so there is a free bug fix.

trieloff:
@tripodsan @kptdobe looks good, but how do we update the smoke tests, so that they accept the new empty alt attributes?

https://app.circleci.com/pipelines/github/adobe/helix-continuous/4031/workflows/25fc08a8-026d-4432-b09d-c255e1c1e4a8/jobs/19076/tests

tripodsan:
@trieloff the smoke tests fail due to another change:

```
> [!NOTE]
> This is a note. Who'd have noted?
```

used to be rendered as:

```
        <div class="admonition-note">
                <p><strong>note</strong>
                        This is a note. Who’d have noted?</p>
        </div>
```

but now it's just:

```
<blockquote>
<p>[!NOTE]
This is a note. Who’d have noted?</p>
</blockquote>
```

trieloff:
@tripodsan we could treat this here as a breaking change and then fix the pre.js example in helix-cli independent from the resolution of https://github.com/micromark/micromark/issues/33

tripodsan:
> @tripodsan we could treat this here as a breaking change and then fix the pre.js example in helix-cli independent from the resolution of [micromark/micromark#33](https://github.com/micromark/micromark/issues/33)

yes. it looks like they will not change it. we'd need to implement our own extension for it. but frankly, I don't know if anyone really needed this.

trieloff:
This specific example was modeled off a thing that the experience cloud docs team used/wanted/needed. 

So we can release pipeline, start using it in pages right away and see how to adapt (or drop) the example in cli. I probably wouldn't build a full-blown extension, just transform the existing MDAST.

trieloff:
@tripodsan https://github.com/adobe/helix-cli/pull/1563 fixes the smoke tests and https://github.com/adobe/helix-pipeline/pull/878/commits/595c3955d2a0a12d20701f297c32301a35c2cf35 flags this as a breaking change.

Downstream projects do not need to increase their major version.

PR #903, opened 2020-12-04
feat: force http1 to avoid issues with long-lived http2 sessions

stefan-guggisberg:
The upcoming/new I/O Runtime environment is Azure-based and supports HTTP/2. This might lead to issues with long-lived HTTP/2 sessions.


trieloff (approved):

src/utils/Downloader.js
```diff
@@ -32,14 +32,11 @@ class Downloader {
     if (!secrets || !secrets.HTTP_TIMEOUT) {
       logger.warn('No HTTP timeout set, risk of denial-of-service');
     }
-    if (options.forceHttp1 || process.env.HELIX_PIPELINE_FORCE_HTTP1) {
```
tripodsan, line 35: keep the `if` and just set the  `process.env.HELIX_PIPELINE_FORCE_HTTP1 = true`

tripodsan (changes requested):
keep the `if` so we can enable it later again

tripodsan (approved):

PR #923, opened 2020-12-29
feat(pipe): migrate to helix-deploy

### LINKED ISSUE #920: migrate to helix-deploy
see adobe/project-helix#508

- [ ] migrate code to use helix-deploy adapter

tripodsan:
fixes #920


trieloff (approved):
In my mind, we would have broken the API and passed Request and Context as-is, but the non-breaking approach is probably better.

trieloff (approved):

tripodsan:
> In my mind, we would have broken the API and passed Request and Context as-is, but the non-breaking approach is probably better.

hmm...yes...but it brings a lot of hack with it.... I think I make it breaking....

tripodsan:
> > In my mind, we would have broken the API and passed Request and Context as-is, but the non-breaking approach is probably better.
> 
> hmm...yes...but it brings a lot of hack with it.... I think I make it breaking....

actually.... keeping it backward compatible makes it easier to maintain (for now). also the _normal_ helix-cli can still operate on the openwhisk version. for helix-pages using helix-deploy will be special....

trieloff (approved):

kptdobe (approved):

rofe (approved):

stefan-guggisberg (approved):

PR #947, opened 2021-01-28
Fix server timing and response headers

### LINKED ISSUE #945: production check doesn't work on AWS
**Description**
the current check for production does:

```
    return Boolean(process.env.__OW_ACTIVATION_ID);
```

which does not work on AWS.

### LINKED ISSUE #946: timing.js sets array as response headers
**Description**
The `server-timing` code sets the response header as array:

```
    context.response.headers['Server-Timing'] = [
      timing,
      `total;dur=${total}`,
    ];
```

which happens to work with node-fetch, but is ignored in helix-fetch, which returns interresting errors on AWS.

tripodsan:
fixes #945 
fixes #946

### DIFF
--- a/src/utils/is-production.js
+++ b/src/utils/is-production.js
@@ -15,7 +15,7 @@
 function production() {
   try {
     // eslint-disable-next-line no-underscore-dangle
-    return Boolean(process.env.__OW_ACTIVATION_ID);
+    return Boolean(process.env.__OW_ACTIVATION_ID) || Boolean(process.env.AWS_LAMBDA_FUNCTION_NAME);
   } catch (e) {
     // this error only occurs when running inside OpenWhisk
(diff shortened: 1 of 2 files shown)


src/utils/is-production.js
```diff
@@ -15,7 +15,7 @@
 function production() {
   try {
     // eslint-disable-next-line no-underscore-dangle
-    return Boolean(process.env.__OW_ACTIVATION_ID);
+    return Boolean(process.env.__OW_ACTIVATION_ID) || Boolean(process.env.AWS_LAMBDA_FUNCTION_NAME);
```
trieloff, line 18: I'd prefer to have something in the wrapper, but it's good for now.

trieloff (approved):

PR #955, opened 2021-02-03
feat(images): do not assume domain root is accessible, use `blob.hlx` URLs

trieloff:
This is changing the default blob URL from `/hlx_…` to `./blob.hlx/…` so that it can work at any level

BREAKING CHANGE: Blob image URL changes from absolute to relative, requires https://github.com/adobe/helix-publish/pull/718


tripodsan (changes requested):
let's wait for the discussion in https://github.com/adobe/helix-publish/pull/718 to be resolved.

trieloff:
Closing, as the URL pattern will be different. Expect an updated PR next week.

PR #992, opened 2021-03-05
fix(html.pipe): add last-modified to response

dominique-pfister:
relates to https://github.com/adobe/helix-content-proxy/pull/306 and https://github.com/adobe/helix-index-pipelines/issues/442

### DIFF
--- a/src/html/fetch-content.js
+++ b/src/html/fetch-content.js
@@ -87,6 +87,12 @@ async function fetchContent(context, {
       sourceLocation,
       sourceHash,
     };
+
+    const lastModified = res.headers.get('last-modified');
+    logger.info(`last-modified: ${lastModified}`);
+    if (lastModified) {
+      context.content.data = { ...context.content.data, lastModified };
+    }
   } catch (err) {
     // throw error if no status error
     /* istanbul ignore next */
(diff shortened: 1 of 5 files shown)


trieloff (approved):

src/html/fetch-content.js
```diff
@@ -87,6 +87,12 @@ async function fetchContent(context, {
       sourceLocation,
       sourceHash,
     };
+
+    const lastModified = res.headers.get('last-modified');
+    logger.info(`last-modified: ${lastModified}`);
+    if (lastModified) {
+      context.content.data = { ...context.content.data, lastModified };
```
tripodsan, line 94: why not: `context.content.data.lastModified = lastmodified` ? this avoids creating a new object.

src/html/set-last-modified.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+const { setdefault } = require('ferrum');
+
+function lastModified(context, { logger }) {
+  const content = setdefault(context, 'content', {});
+  const res = setdefault(context, 'response', {});
+  const headers = setdefault(res, 'headers', {});
+
+  // somebody already set a last-modified key
+  if (headers['Last-Modified']) {
```
tripodsan, line 20: I think it's ok for now, but we should probably use a case insensitive map here. maybe the headers class from helix-fetch....

tripodsan (approved):

PR #1037, opened 2021-04-09
fix: do not cache responses to Fastly content-proxy

dominique-pfister:
When requesting a resource from Fastly content-proxy, do not cache the response.

### DIFF
--- a/src/html/fetch-content.js
+++ b/src/html/fetch-content.js
@@ -61,6 +61,7 @@ async function fetchContent(context, {
       options: {
         timeout: secrets.HTTP_TIMEOUT_EXTERNAL,
         headers,
+        cache: 'no-store',
       },
       errorOn404: false,
     });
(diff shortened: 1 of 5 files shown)


trieloff (approved):
@stefan-guggisberg ?

stefan-guggisberg (approved):

src/html/fetch-content.js
```diff
@@ -55,7 +55,7 @@ async function fetchContent(context, {
   }
 
   try {
-    logger.debug(`loading from content proxy: ${contentProxyUrl}`);
+    logger.info(`loading from content proxy: ${contentProxyUrl}`);
```
dominique-pfister, line 58: Note, for easier debugging, it would be great to know the content proxy URL - is that ok with everybody, it hopefully does not reveal anything it shouldn't.

trieloff (approved):

